PR #2, opened 2018-04-01
jsonp endpoint

radius314:


radius314:
@jbraswell is this implemented in tomato?

LittleGreenViper:
I merged it in, as I feel it's safe enough. It doesn't touch any of the existing code. However, we should give it good testing, and add it to the documentation.

radius314:
ACK

radius314:
LGTM https://codepen.io/radius314/pen/RMxQrB

LittleGreenViper:
Sweet!
I have also sent the code to the test server: https://tkddevel.com/bmltwork/bmlt/main_server/

jbraswell:
@radius314 jsonp is unsupported in tomato

LittleGreenViper:
I've never been that thrilled with JSONP. It seems a bit of a hack. However, I'm happy to support it in this fashion.
We should remember that whatever we add, we add forever. We'll be responsible for supporting it until The End Times.
Microsoft is even known for PUTTING BUGS BACK in their software, as developers had coded their apps to compensate for the issues, and the fixes broke apps.

radius314:
Yeah it’s not ideal.  But it’s a lot easier than explaining CORS headers.
It’s a simple feature to support forever.

Issue #6, opened 2018-05-04
BMLT.JS AM/PM


pjaudiomv:
when an hour is single digit in the AM it displays a leading zero, would be nice to chop this off.

Ex. 

Primary Purpose Group
(O,D,SD,WC)
Royal Baptist Church
4761 Luella Avenue
North Charleston, SC
Saturday **09:00 AM**

this only seems to happen in the AM
found this but didnt seem to help me figure it out

https://stackoverflow.com/questions/29206453/best-way-to-convert-military-time-to-standard-time-in-javascript?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

@radius314

pjaudiomv:
![bmlt](https://user-images.githubusercontent.com/34245618/40282004-60d85ff4-5c36-11e8-86e5-1a931ce46817.png)

PR #7, opened 2018-05-04
fix for leading zeros on bmlt.js

pjaudiomv:
Fixing #6

### DIFF
--- a/main_server/local_server/bmlt.js
+++ b/main_server/local_server/bmlt.js
@@ -17,26 +17,26 @@ var militaryToStandard = function(value) {
         }
         else {
             if(value.length == 8){ //If value is the expected length for military time then process to standard time.
-				value = value.split(':'); // convert to array
-				// fetch
-				var hours = Number(value[0]);
-				var minutes = Number(value[1]);
+                value = value.split(':'); // convert to array
+                // fetch
+                var hours = Number(value[0]);
+                var minutes = Number(value[1]);
 
-				// calculate
-				var timeValue;
-				if (hours > 0 && hours <= 12) {
-					timeValue= "" + hours;
-				} else if (hours > 12) {
-					timeValue= "" + (hours - 12);
-				} else if (hours == 0) {
-					timeValue= "12";
-				}
- 
-				timeValue += (minutes < 10) ? ":0" + minutes : ":" + minutes;  // get minutes
-				timeValue += (hours >= 12) ? " PM" : " AM";  // get AM/PM
-				// show
-				return timeValue;
-				}
+                // calculate
+                var timeValue;
+                if (hours > 0 && hours <= 12) {
+                    timeValue= "" + hours;
+                } else if (hours > 12) {
+                    timeValue= "" + (hours - 12);
+                } else if (hours == 0) {
+                    timeValue= "12";
+                }
+                
+                timeValue += (minutes < 10) ? ":0" + minutes : ":" + minutes;  // get minutes
+                timeValue += (hours >= 12) ? " PM" : " AM";  // get AM/PM
+                // show
+                return timeValue;
+                }
             else { //If value is not the expected length than just return the value as is
                 return value;
             }
--- a/main_server/local_server/bmlt.js
+++ b/main_server/local_server/bmlt.js
@@ -17,9 +17,9 @@ var militaryToStandard = function(value) {
         }
         else {
             if (value.length == 8) { //If value is the expected length for military time then process to standard time.
-                value = value.split(':'); // convert to array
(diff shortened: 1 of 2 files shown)


jbraswell:
Looks like you’re using tabs instead of spaces for indentation. Will you switch to spaces so that your changes match the rest of the file?

LittleGreenViper:
Looks OK to me. This isn't really my code, anyway.

I tend to use spaces, but I do it by setting my editor to interpret tabs as as spaces, so sometimes, there may be some tabs in there.

pjaudiomv:
yea after jonathan said that i changed my editor to interpret tabs as spaces as well

LittleGreenViper:
Johnathan's a Python programmer.

Spaces are very important in Python.

pjaudiomv:
yes hes a snake wrangler

radius314:
Ok I see that this addressing #6.  Looking at it again with this new knowledge.

main_server/local_server/bmlt.js
```diff
@@ -16,22 +16,26 @@ var militaryToStandard = function(value) {
...
         else {
-            if(value.length == 8){ //If value is the expected length for military time then process to standard time.
-                var hour = value.substring ( 0,2 ); //Extract hour
-                var minutes = value.substring ( 3,5 ); //Extract minutes
-                var identifier = 'AM'; //Initialize AM PM identifier
+            if (value.length == 8) { //If value is the expected length for military time then process to standard time.
+                value = value.split(':'); // convert to array
+                // fetch
+                var hours = Number(value[0]);
+                var minutes = Number(value[1]);
```
radius314: Since you are not manipulating minutes, why not just leave it a string, instead of re-converting to a string on line 35?

main_server/local_server/bmlt.js
```diff
@@ -32,7 +31,7 @@ var militaryToStandard = function(value) {
                     timeValue= "12";
                 }
                 
-                timeValue += (minutes < 10) ? ":0" + minutes : ":" + minutes;  // get minutes
```
radius314: Why is this line even necessary?

pjaudiomv:
you mean something like this?

pjaudiomv:
sorry just finishing up at my asc, I also must stress as I have many times before I am not a programmer so whatever I do you guys will most likely always be right.

main_server/local_server/bmlt.js
```diff
@@ -16,22 +16,25 @@ var militaryToStandard = function(value) {
...
-                    hour=12;
-                }
-                if(hour > 12){ //If hour is greater than 12 then convert to standard 12 hour format and set the AM PM identifier to PM
-                    hour = hour - 12;
-                    identifier='PM';
-                }
-                return hour + ':' + minutes + ' ' + identifier; //Return the constructed standard time
+                
+                timeValue += (Number(value[1]) < 10) ? ":0" + Number(value[1]) : ":" + Number(value[1]);  // get minutes
+                timeValue += (hours >= 12) ? " PM" : " AM";  // get AM/PM
```
radius314: Swap line 34 for

`timeValue += value[1]`

main_server/local_server/bmlt.js
```diff
@@ -16,22 +16,25 @@ var militaryToStandard = function(value) {
             return value;
         }
         else {
-            if(value.length == 8){ //If value is the expected length for military time then process to standard time.
-                var hour = value.substring ( 0,2 ); //Extract hour
-                var minutes = value.substring ( 3,5 ); //Extract minutes
-                var identifier = 'AM'; //Initialize AM PM identifier
+            if (value.length == 8) { //If value is the expected length for military time then process to standard time.
+                value = value.split(':'); // convert to array
```
radius314: One other suggestion is to use a different variable other than `value` for assignment.  It's generally not a good idea to overwrite variables.  You lose access to the original value, and you make yourself prone to other such errors potentially.

main_server/local_server/bmlt.js
pjaudiomv: I think we would have to do something like this to get the seperator ':'
":" + value[1]

radius314: Yes, my initial comment was revised.

main_server/local_server/bmlt.js
pjaudiomv: makes sense

radius314:
@pjaudiomv I'm ready to merge, can you update the CHANGELIST section of README.md?

pjaudiomv:
oh dear what'd I do

radius314:
You need to rebase from master, I resolved the conflict here.

PR #11, opened 2018-05-26
*DO NOT MERGE* auto save geocode

radius314:
**DO NOT MERGE**

This pull request needs some discussion, it's definitely not ready to be merged. 

In working this issue a couple of things comes to mind.  

1) Not everyone will want to do this.
2) People will want granular control.
3) This will not help mobile apps, so perhaps this should be serverside.

Looking for commentary, feedback before I spend any more time on this.

@jbraswell @pjaudiomv @LittleGreenViper

### DIFF
--- a/main_server/local_server/server_admin/server_admin_javascript.js
+++ b/main_server/local_server/server_admin/server_admin_javascript.js
@@ -1439,21 +1439,27 @@ function BMLT_Server_Admin ()
                         {
                         save_a.className = 'item_hidden';
                         throbber_span.className = 'bmlt_admin_ajax_button_throbber_span';
-                        this.sendMeetingToServer ( in_meeting_id, false );
-                        meeting_sent = true;
+                        self = this;
+                        this.setMapToAddress( in_meeting_id, function(error)
+                            {
+                            self.sendMeetingToServer ( in_meeting_id, false );
+                            meeting_sent = true;
+
+                                if ( c == this.m_search_results - 1 && !meeting_sent )
+                                {
+                                    save_a.className = 'item_hidden';
+                                    throbber_span.className = 'bmlt_admin_ajax_button_throbber_span';
+                                    self.setMapToAddress( in_meeting_id, function(error) {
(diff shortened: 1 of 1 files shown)


pjaudiomv:
cool!
for numbers
1. yes this should be option
2. yes granular control is key as even if location is entered correct google doesnt always geocode things properly.
3. i think server side would be best way to go as well

LittleGreenViper:
Oh...yeah, one other thing. Patrick is right. This should be done server-side, in PHP.

radius314:
Ok, the server side part is definitely what I was thinking, in order to properly encapsulate the feature.

It seems as the default function should be to always do the lookup, and then provide the ability to handle the exception cases mentioned in 1 + 2 of your comment @LittleGreenViper.

I can't think of an easy to create this way to override the proposed auto-save feature.

LittleGreenViper:
It's a great idea, but nuking people's custom settings will cause insane rage from insane ragers.

This was something that I gave a lot of thought to as I set things up. I had to hode by dose (hold my nose) when I did things the way I did. I hate to compromise user experience; especially when it increases the chance for error.

radius314:
I know what you mean and I'm cautious of that as well...

Maybe the approach should be an option in your config file to turn this on.  It's not the default experience so people can opt-in or choose to keep this feature status quo.

radius314:
The one part that would be missing would be a way to allow individual overrides.  Someone will want the best of both worlds... I can think of a way to hack it.  Perhaps 

When you say the installer needs work, do you have more detail about what work?

LittleGreenViper:
I think that I have added one or two fields (can't remember what, exactly) that aren't reflected in the GUI. These are usually checkboxes.

It's a big fat pain to tweak the GUI. It's not the cleanest JS on Earth. I kinda hacked it out over a weekend, because I was getting a lot of complaints about setting up the config file.

As far as documenting the file, it would probably be enough to just add some good comments into a template file, like what I do with the BMLT-Basic config file.

PR #13, opened 2018-07-25
add sb contact email to semantics

pjaudiomv:
wondering if it would be possible to add the service body contact info to the semantic responses. If for some reason there's a problem with this. is there any other way to get this data through the API


pjaudiomv:
holy crap, and a bunch more explecitives i am blind my bad

pjaudiomv:
g_include_service_body_email_in_semantic = TRUE;

pjaudiomv:
it was right there in front of my face like many times

PR #21, opened 2018-09-28
cache busting

pjaudiomv:


radius314:
Why cache busting json2?  It shouldn’t change

pjaudiomv:
I actually think I should just revert and cancel this pr. as were actually calling the js_stripper.php file with the js filename querystring we would need to adjust the preg_match in js_stripper.php or use headers instead. I need to get better with regex but right now im not.

radius314:
regex101.com

radius314:
Don’t feel too bad, 21 years of writing code and i still haven’t memorized
all the regex syntax.

pjaudiomv:
yea that site is awesome, good tool to know.

LittleGreenViper:
Just FYI: The js_stripper and style_stripper files are becoming problematic (and less necessary).

WordPress (for one) is now no longer allowing execution of files inside the directory (good security policy), so these files often result in broken JS and styles.

Remember that the BMLT is ten years old. That's Methuselah in Internets years. When I originally wrote it, it was important to optimize pages, and my code is HEAVILY commented (a habit that is sadly out of vogue, these days). I COMPLETELY disagree with just not commenting files. If the meta causes issues in load times, we need to find different ways to optimize.

Many sites don't bother to optimize their CSS and JS. If you aren't Google, getting hit a billion times an hour, then the optimization doesn't actually buy much.

Much as we like to think we're all that, in reality, NA sites barely make the bandwidth meter twitch. The entire TOMATO database is a rounding error in Big Data World.

radius314:
Yeah, I agree.  Also, JS and CSS optimization require more advanced tooling
which adds an additional set of skills in terms of getting other people to
be involved in a project.

I was considering minifiying and uglifying crouton JSS using webpack and
decided it wasn't worth the effort or locking out people that don't know
about it.

PR #23, opened 2018-09-29
Fix array2json to handle integers and booleans properly

jbraswell:
This fixes an issue where the user is warned that they will lose changes when switching between users and service bodies, even after they have properly saved their changes. The problem was that the json returned by the server was not properly encoded, so that comparisons between the object in browser memory and the returned saved in the javascript callback handlers were failing.

I validated that saving Users, Service Bodies, and Formats now works properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ For instructions on installing the root server, see [the page on installing a ne
 CHANGELIST
 ----------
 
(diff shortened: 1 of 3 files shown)


radius314:
Glad to see this bug fixed.

PR #24, opened 2018-09-29
Handle cuts and pastes properly when editing objects

jbraswell:
This addresses https://github.com/LittleGreenViper/BMLT-Root-Server/issues/20.

The `onpaste` and `oncut` events are fired _before_ the cut or paste actually takes place. To work around this, we wrap things in setTimeout.

Some of these handlers were just calling the wrong functions. Other inputs didn't have handlers at all.


jbraswell:
The timeout time of 0 means put it on the JavaScript event loop queue immediately, which puts it behind the actual cutting or pasting (which is already in that queue), making sure the text has been pasted before the handler is called. 

Agreed that this one needs some good browser testing. I think we should hit all the modern browsers. Thank God IE6 is no longer supported by MS.

radius314:
It’d good to declare bankruptcy on old browsers and use whatever Wordpress
standards are.

Speaking of which PHP 5.6 is EOL at end of the year.

PR #33, opened 2018-10-06
 add meeting_states_and_provinces to GetServerInfo 

pjaudiomv:


LittleGreenViper:
Lemme check...

LittleGreenViper:
Damn. Good job, Patrick! He got it!

pjaudiomv:
holy hole in a donut batman, ill remember this day. A PR I didn't need to revise!

jbraswell:
Ohhhh, now I see what this is for! Thanks Patrick

pjaudiomv:
i decided not to slash out double quotes. 1. as you said its a controlled set of values and not likely to be an issue. but mainly because of 2. we are converting special characters to html entities already so if I did that and for some crazy reason someone added a double quote to the meeting_states _and_provinces array you would end up with something like this \&quot; instead of &quot  this happens here https://github.com/LittleGreenViper/BMLT-Root-Server/blob/master/main_server/client_interface/csv/csv.php#L475

jbraswell:
Does PHP have a built in function for escaping xml?

pjaudiomv:
htmlspecialchars is what i would use. http://php.net/htmlspecialchars

pjaudiomv:
or you use DOM to build your xml

pjaudiomv:
I realize this is low priority but bare with me here as im amateur and always trying to learn best practices, luckily i have some good teachers thanks to you guys. escaping a double quote with csv can be tricky in this scenario because the meeting_states _and_provinces string is already escaped using double quotes as it contains commas if we did this in the GetServerInfo() function we would end up with erroneous &quot; in the xml and json responses. I would still not deal with this as I don't know any states or provinces that use a double quote, single yes. we also can't deal with this in the GetServerInfo switch statement easily because then were working with the whole file.  My suggestion is if we did want to deal with this is just to replace a double quote with its htmlentity &quot;, this would be fine for all semantic responses and just leave a &quot; in the csv output. we already do this with the GetChanges csv response.

LittleGreenViper:
The double-quotes (or single-quotes) used to define the Array are not reflected in the actual values.

i.e:

$some_array_of_strings = Array("String 1", "String 2", "String 3");

is actually:

0: String 1
1: String 2
2: String 3

Internally, and when you do an implode:

$some_string = implode(",", $some_array_of_strings);

You get:

"String 1, String 2, String 3"

But if you do:

$some_array_of_strings = Array('"String 1"', '"String 2"', '"String 3"');

or:

$some_array_of_strings = Array("\"String 1\"", "\"String 2\"", "\"String 3\"");

Then you would get:

"String 1", "String 2", "String 3"

LittleGreenViper:
Good point.

Moot, however, as we have a controlled data input. Just make sure that you know what bad data looks like, so you don't get caught up trying to help someone that accidentally causes it.

pjaudiomv:
gotcha thanks, and thanks for your patience.

PR #37, opened 2018-11-04
Fix bad default datetime, which can cause migration from 2.11.0 to fail

jbraswell:


jbraswell:
I've seen this on two root servers in the wild.

PR #38, opened 2018-11-06
Automatically calculate coordinates when saving meeting...

jbraswell:
* Only geocodes when address has changed
* Removes set lat/long button
* Location map still exists to allow tweaking
* Displays a giant WARNING in the admin user interface when the google maps api key is invalid, and links to a blog post about acquiring an api key.
* Displays an alert, and will not save a meeting that requires geocoding when the api key is invalid.
* Install wizard was modified...
  * Removed map for setting default coordinates, because it doesn't make sense when we geocode every meeting.
  * Added a "TEST KEY" button.
  * Display a big red warning message on the final page of the wizard if the API key is not set or is invalid.


radius314:
Is there going to be a mechanism to allow manually set coordinates?

jbraswell:
At this point in the PR, the only way to manually set coordinates is through the location map. May need to add a manual override but I’m not yet convinced it’s necessary.

radius314:
can you just have two separate buttons?  update but not adjust coordinates?

radius314:
the location map just updates the lat/long text fields doesn't it?

jbraswell:
I don’t like the idea of having two separate buttons, but it’s certainly possible. 

Yeah, the location map just updates lat/long fields.

jbraswell:
Because google api keys are now pretty much required to be able to edit meetings, the following work should be done before this feature is completed:

1. Modify the installation wizard to require a valid google api key, and link a tutorial explaining how to acquire one.
2. Display an error message in the admin user interface when the google api key is invalid, and link a tutorial explaining how to acquire one.
3. Create said tutorial.

radius314:
I wonder if it makes sense to be able to report this on the serverInfo.

jbraswell:
By report this, do you mean the validity of the server’s API key? That could be interesting information to have as part of server info. Don’t think it’s absolutely necessary for this feature, but it could help us find servers without valid API keys and be proactive about contacting admins.

radius314:
exactly

radius314:
Ok, so far my testing looks good.  From what I can tell, the location map doesn't allow you to manually set coordinates.  It allows you to find a location and then use that to set an address.

I'm not sure how many folks are manually setting the lat/long.  By this, meaning there is a different lat/long from the actual "addressed" location.

radius314:
Poll it on fb group?

jbraswell:
Outstanding TODOs:

1. Modify the installation wizard to require a valid google api key, and link a tutorial explaining how to acquire one.
2. Potentially add a "Manually Override Coordinates" checkbox, which enables the latitude and longitude inputs, allowing users to manually enter coordinates. If we do this, we'll likely need to persist whether or the coordinates were overridden alongside the meeting information.

radius314:
Before embarking on Item 2 my suggestion would be to poll the community and see how many people actually need this feature, and why.

jbraswell:
I *think* this is done. Needs testing.

Issue #57, opened 2018-11-20
Multiple formats may be assigned to the same NAWS format, but the NAWS Dump doesn't respect this


jbraswell:
Lori noticed this when going through the NAWS Dump of the South Florida region. Every single meeting was listed as Closed. This is because multiple formats in the server are mapped to 'OPEN' (because it's really several servers merged together by lettuce), and the data structure used by the GetNAWSDump code doesn't allow this. Basically it ends up using the BMLT Format -> NAWS Format mapping for the BMLT Format stored last in the table.

I'm going to need to fix this imminently to get exports for merged servers working.

PR #59, opened 2018-11-22
Support multiple formats being assigned to the same NAWS Format ID

jbraswell:
This fixes https://github.com/bmlt-enabled/BMLT-Root-Server/issues/57

This is a bug that only impacts servers that were merged together by lettuce. The only server I'm aware of that would be impacted by this is the SEZF server. I've already patched the SEZF server with this code.

The issue is that after merging with lettuce, multiple formats can be mapped to the same NAWS Format ID. The mapping dictionary didn't allow for this. Fixed it by storing an array of matching format IDs at each world ID key in the `$g_format_dictionary`, and then fixing all access of that dictionary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ For instructions on installing the root server, see [the page on installing a ne
 CHANGELIST
 ----------
 
+***Version 2.12.2* ** *- TBD*
+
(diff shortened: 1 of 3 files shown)


radius314:
Any required change to lettuce?

jbraswell:
No changes are required to lettuce. The data was fine, the naws dump just didn't support this until now.

Issue #63, opened 2018-11-26
Request - Alphabetize Service Body checkboxes on Meeting Search page


Redguy1959:
This will enhance useability.

pjaudiomv:
this is something that has annoyed me as well, just put in a PR to fix #66 , well see how that pans out.

jbraswell:
This is fixed, and will be in the next release. Thanks @pjaudiomv

PR #66, opened 2018-11-26
sort checkboxes in advanced service bodies section #63

pjaudiomv:


jbraswell:
Did you test this change? I think this array contains objects, not strings, so you may not get the result you expect.

jbraswell:
You might need to do something like https://stackoverflow.com/questions/4282413/sort-array-of-objects-by-object-fields instead.

pjaudiomv:
I did test it, and worked fine on my test server.

jbraswell:
Cool. I still don't think this will always work. I think by default the sort function does a comparison on the first member of the object, which isn't guaranteed. Probably using usort with a comparison function is the way to go.

pjaudiomv:
ok cool, ill check that out

pjaudiomv:
I looked into it and im doing the exact same thing for the other sorting methods used to sort the service bodies in dropdowns.

https://github.com/radius314/BMLT-Root-Server/blob/40fd1b894546c1a575152f060f47960cc5e157fa/main_server/local_server/server_admin/c_comdef_admin_main_console.class.php#L982

https://github.com/radius314/BMLT-Root-Server/blob/40fd1b894546c1a575152f060f47960cc5e157fa/main_server/local_server/server_admin/c_comdef_admin_main_console.class.php#L1619

pjaudiomv:
I did a var dump on all of those and they all essentially output same data format

pjaudiomv:
they're arrays not objects

jbraswell:
Yes, arrays of what?

LittleGreenViper:
The sort is an array function, and it operates by comparing whatever the array aggregates.

For example:

    $pooch = Array(1,3,2);
    sort($pooch);

makes $pooch come out to 1,2,3

However, if it is:

    class Screw {
        var number;
        __construct($inNumber) {
            $this->number = $inNumber;
        }

        $pooch = Array(Screw(1), Screw(3), Screw(2));
        sort($pooch);

$pooch is undefined. I suspect the sort simply fails.

Forgive me if this is stupid PHP. I've been out of PHP head for a couple of months.

pjaudiomv:
I can't seem to figure this out, out of my skill set / understanding

PR #68, opened 2018-11-27
use usort for both service bodies and users, create function for sorting

pjaudiomv:
this could most likely just be done like so, just sort the users and service bodies once


pjaudiomv:
I have tested this in every instance in which its used, for service body dropdowns, user dropdowns, user checkbox lists on service body user assignments, service body checkbox list on meeting search

Issue #69, opened 2018-11-27
default_duration_time from autoconfig doesnt get respected in meeting editor


pjaudiomv:
the default_duration_time from autoconfig doesnt get respected in meeting editor dropdown, im guessing it should? at least that would make since to me. the default dropdown value is a static pulled from the server admin strings. I feel like somehow we should pull from $localized_strings['default_duration_time']. im using latest version 2.12.2

maybe thats done here
https://github.com/bmlt-enabled/BMLT-Root-Server/blob/a9da7caeb9ad31494d0e4ca4f39aae27075669bf/main_server/local_server/server_admin/c_comdef_admin_main_console.class.php#L349

this is the static its currently getting pulled from
https://github.com/bmlt-enabled/BMLT-Root-Server/blob/a9da7caeb9ad31494d0e4ca4f39aae27075669bf/main_server/local_server/server_admin/lang/en/server_admin_strings.inc.php#L139

jbraswell:
I think you'd want to fix this code in the dom template that gets printed here:

https://github.com/bmlt-enabled/BMLT-Root-Server/blob/a9da7caeb9ad31494d0e4ca4f39aae27075669bf/main_server/local_server/server_admin/c_comdef_admin_main_console.class.php#L1659-L1679

jbraswell:
Basically you'd just add a "selected" attribute to the hour and minute options if they match the default.

`<option value="blar" selected>asdf</option>`

jbraswell:
fixed

PR #70, opened 2018-11-27
Fix for default duration time not displaying correctly on meeting edi…

pjaudiomv:
…tor #69 

one issue issue with this is it will never use 'comdef_server_admin_strings']['meeting_editor_default_duration'] which may not be an issue if default_duration_time is not set it will default to 1:00 per

https://github.com/bmlt-enabled/BMLT-Root-Server/blob/a9da7caeb9ad31494d0e4ca4f39aae27075669bf/main_server/server/c_comdef_server.class.php#L2900

maybe itsead of just defaulting to 1:00 it should then default to the language meeting_editor_default_duration

c_comdef_server::GetLocalStrings()['comdef_server_admin_strings']['meeting_editor_default_duration']


radius314:
does this work on IE11? 😂

pjaudiomv:
omg that reminds me, crouton.

pjaudiomv:
there this should do it, if someone doesn't have default_duration_time set in there autoconfig it will default to the meeting_editor_default_duration in comdef_server_admin_strings. If they don't have that set in there lang file, then well they should  set it and it then defaults to 0.

main_server/local_server/server_admin/c_comdef_admin_main_console.class.php
```diff
@@ -346,7 +346,7 @@ function return_main_console_html()
                 $ret .= 'var g_Save_meeting_button_name = \''.self::js_html ( $this->my_localized_strings['comdef_server_admin_strings']['meeting_save_buttonName'] ).'\';'.(defined ( '__DEBUG_MODE__' ) ? "\n" : '');
                 $ret .= 'var g_default_meeting_weekday = '.intVal ( $this->my_localized_strings['comdef_server_admin_strings']['meeting_editor_default_weekday'] ).';'.(defined ( '__DEBUG_MODE__' ) ? "\n" : '');
                 $ret .= 'var g_default_meeting_start_time = \''.self::js_html ( $this->my_localized_strings['comdef_server_admin_strings']['meeting_editor_default_start_time'] ).'\';'.(defined ( '__DEBUG_MODE__' ) ? "\n" : '');
-                $ret .= 'var g_default_meeting_duration = \''.self::js_html ( $this->my_localized_strings['comdef_server_admin_strings']['meeting_editor_default_duration'] ).'\';'.(defined ( '__DEBUG_MODE__' ) ? "\n" : '');
+                $ret .= 'var g_default_meeting_duration = \''.self::js_html ( c_comdef_server::GetLocalStrings()['default_duration_time'] ).'\';'.(defined ( '__DEBUG_MODE__' ) ? "\n" : '');
```
jbraswell: Would `$this->my_localized_strings['comdef_server_admin_strings']['default_duration_time']` have worked yere?

jbraswell:
Have you tested this code?

main_server/local_server/server_admin/c_comdef_admin_main_console.class.php
pjaudiomv: no that would not of worked, because we are inside that function

pjaudiomv:
I have tested it with and without the default_duration_time set in the autoconfig and works as expected

main_server/local_server/server_admin/c_comdef_admin_main_console.class.php
jbraswell: Can you explain why you didn't just fix the setting of `meeting_editor_default_duration` so that it would always contain a sane default? This seems like a bit of indirection, having a default value stored in two different places, with possibly different values.

jbraswell: Seems like `$this->my_localized_strings['comdef_server_admin_strings']['default_duration_time']` should exist, and if it doesn't, it's probably a problem with this line:

https://github.com/bmlt-enabled/BMLT-Root-Server/blob/master/main_server/server/c_comdef_server.class.php#L2900

pjaudiomv: yes so if the `default_duration_time` isn't set, it was hardcoded in c_comdef_server.class.php to be 01:00:00 however this was only accessed in rare circumstances as the default was set to `meeting_editor_default_duration` and not to `default_duration_time`. I felt instead of being hardcoded it should just pull from `meeting_editor_default_duration`

pjaudiomv: yes it should exist, but if it doesnt theyre should be a fallback default no?

jbraswell: But now you're pulling from `default_duration_time`, not `meeting_editor_default_duration`. And you're pulling in a manner that is syntactically inconsistent with the lines above and below.

pjaudiomv: oh wait hold on i was confused by your last statement

pjaudiomv: cool, damn i greped it earlier too. should of known that

pjaudiomv: > In PHPStorm you can do that kind of searching with shift+ctrl+f

Um, its shift+apple+f...

pjaudiomv: wait so should I then have no fallback if  default_duration isnt set in the autoconfig anymore, or still set it and keep it at 01:00:00 or something.

pjaudiomv: it kind of has to be setup, the only way i see it not being set is if you remove it

pjaudiomv: as its set in the installer wizard

jbraswell: I think keeping the default at 01:00:00 is fine

pjaudiomv: I just don't like the idea of hardcoding the fallback default,however region bias is also hardcoded to us. so guess its not that big of a deal either way.

jbraswell: Yeah I mean settings have to fallback to something. Either they fallback to something hardcoded or we make the software fail when it's not set. Those are basically your options.

pjaudiomv: I would still rather fall back to `meeting_editor_default_duration` from set lang file `server_admin_strings.inc.php` instead of hardcoded in `c_comdef_admin_main_console.class.php` whats the issue with doing this?

pjaudiomv: im just trying to get better understanding, could this fail at some point. or is the point to not pull extra resources than needed

pjaudiomv: can I talk to my other sponsors till i get the answer I want to here?

pjaudiomv: no that makes sense. language file should be used for language stuff and wording not settings.

pjaudiomv: however there are ALOT of settings in those files

pjaudiomv: actually i shouldnt say a lot, about 6

jbraswell: Yeah, I'd just move them as we have a reason to. No reason to make a risky change unless there's a reason, like fixing this particular issue with this setting.

pjaudiomv: bam, like it never existed

jbraswell:
As long as these latest changes have been tested... :shipit:

pjaudiomv:
nice! yep send it.

PR #71, opened 2018-11-28
auto packaging on tag

radius314:


jbraswell:
Basically the contents need to mirror BMLT-Root-Server.zip in the root of the repo. Then we can just delete that zip from source control, and update the references to it on bmlt.app.

jbraswell:
I’m happy to continue this work if you don’t have the time. My last day at my job is Thursday, and then I have two weeks to work on whatever.

radius314:
just pushed a change that will do as such

radius314:
btw I may need to do a series of tweaks here... I didn't go as far as to test this completely on my fork.

jbraswell:
Does merging this make testing any easier for you, or are you fine working through issues in your fork?

radius314:
yeah i need to merge

radius314:
i can try to work in a branch on this main repo

radius314:
would be good to have an initial sense of pass/fail though.

jbraswell:
Cool. My preference is to work in a non-master branch, but I’m also fine with just merging to master as long as the work is pushed to completion (I know it will be). Use your best judgment. As an owner you can merge PRs on this repo, right?

PR #75, opened 2018-11-28
changing license model

radius314:


LittleGreenViper:
There. I probably just showed up in RMS' crosshairs...

Issue #86, opened 2018-12-02
satellite base class static files, 404


jbraswell:
2(index):12 GET https://bmlt.sezf.org/new_main_server/client_interface/html/bmlt-basic/BMLT-Satellite-Base-Class/themes/BlueAndWhite/style_stripper.php?filename=nouveau_map_styles.css net::ERR_ABORTED 404 (Not Found)
2(index):20 GET https://bmlt.sezf.org/new_main_server/client_interface/html/bmlt-basic/BMLT-Satellite-Base-Class/nouveau_map_search.js net::ERR_ABORTED 404 (Not Found)

pjaudiomv:
didn't realize you self assigned

PR #87, opened 2018-12-02
fix for satellite base class static files #86

pjaudiomv:


jbraswell:
fixes https://github.com/bmlt-enabled/bmlt-root-server/issues/86

PR #91, opened 2018-12-03
redundant span class bmlt_admin_value_left

pjaudiomv:
this causes the input boxes to not line up properly and is redundant. 

screen shot 2018-12-03 at 4 39 29 pm


pjaudiomv:
note that this only happens when not using meeting_states_and_provinces dropdown

PR #102, opened 2018-12-05
use supperior zip for geocode test

pjaudiomv:
this should fix gkey checking issue, more superior zip was needed for checking.


jbraswell:
wat?

pjaudiomv:
My bad hoping you wouldn't check :(

PR #104, opened 2018-12-05
 remove set map to address button

pjaudiomv:


jbraswell:
You also need to remove the `setMapToAddress` function in server_admin_javascript.js. Then you need to remove javascript references to the button.

Lines 2378 and 2397.

pjaudiomv:
good call

README.md
```diff
@@ -70,6 +70,7 @@ CHANGELIST
 - Latitude and longitude input boxes are now read only instead of disabled, allowing users to select and copy the calculated values.
 - Fixed an issue where static files in the client_interface/html/ endpoint were not being included.
 - The German translations for the administration UI were updated.
+- Removed "Set Map to Address button" in the single meeting editor as not needed anymore and messed up workflow.
```
jbraswell: The "Set Map to Address" button in the single meeting editor was removed, as it no longer provides values to the workflow.

pjaudiomv: profresh

PR #105, opened 2018-12-07
update composer for latest sat base class

pjaudiomv:


pjaudiomv:
I did composer update, was i supposed to do anything else?

jbraswell:
This is perfect. Make sure you test that the html endpoint UI works with this version of the base class and then we’re good.

pjaudiomv:
works fine, like a glove

jbraswell:
Make like a tree and get out of here

PR #107, opened 2018-12-08
use case-insensitive sorting

pjaudiomv:
could use this or [strnatcasecmp](http://php.net/strnatcasecmp)  which uses a [natural sort order](https://github.com/sourcefrog/natsort), [strcasecmp](http://php.net/manual/en/function.strcasecmp.php) is probably fine though. this will sort the users and service bodies in admin case-insensitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ For instructions on installing the root server, see [the page on installing a ne
 CHANGELIST
 ----------
 
+***Version 2.12.4* ** *- UNRELEASED*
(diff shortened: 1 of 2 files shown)


jbraswell:
I think we ought to use `strnatcasecmp`

pjaudiomv:
ok yea probably better decision

jbraswell:
have you tested this?

pjaudiomv:
yup mo better

PR #118, opened 2018-12-16
change semantic source url

pjaudiomv:


pjaudiomv:
only on the semantic workshop, at the bottom

pjaudiomv:
we could also choose to just get rid of it

jbraswell:
I think we should just get rid of the link.

Issue #119, opened 2018-12-16
Request - Show logged in user at top of page.


californiasteve:
I would like to see the user at the top of the page. Us with multiple logins it would be nice to see which one I'm logged in with and not have to log out then back in to make sure.

jbraswell:
This makes sense to me. Maybe just change the sign out link to he “Sign Out (<username>)”?

pjaudiomv:
I think having the username in sign out link is great idea.

pjaudiomv:
this is in master and will be in next release

PR #120, opened 2018-12-16
Added user name to sign out link for the administration UI

pjaudiomv:
ref feature request #119

### DIFF
--- a/main_server/local_server/server_admin/c_comdef_login.php
+++ b/main_server/local_server/server_admin/c_comdef_login.php
@@ -148,7 +148,7 @@ if (isset($_SESSION[$admin_session_name])) {
     }
 
     if (!isset($supress_header) || !$supress_header) {
-        echo '<div class="bmlt_admin_logout_bar"><h4><a href="'.$_SERVER['PHP_SELF'].'?admin_action=logout">'.c_comdef_htmlspecialchars($localized_strings['comdef_server_admin_strings']['logout']).'</a></h4>';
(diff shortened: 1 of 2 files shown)


jbraswell:
It’s up to you. If you tested locally and are sure your change works then there’s no need.

pjaudiomv:
cool yea no its good to go

pjaudiomv:
we don't really have any codeline policy and i guess you and chris would be the defacto codeline curators. but makes sense, that last one i wanted top go to unstable because i wanted to test. but this is pretty straight forward.

jbraswell:
All I care about is never merging untested code to master. I don’t care if it’s tested locally or in the unstable branch as long as it’s tested.

main_server/local_server/server_admin/c_comdef_login.php
```diff
@@ -148,7 +148,7 @@
     }
 
     if (!isset($supress_header) || !$supress_header) {
-        echo '<div class="bmlt_admin_logout_bar"><h4><a href="'.$_SERVER['PHP_SELF'].'?admin_action=logout">'.c_comdef_htmlspecialchars($localized_strings['comdef_server_admin_strings']['logout']).'</a></h4>';
+        echo '<div class="bmlt_admin_logout_bar"><h4><a href="'.$_SERVER['PHP_SELF'].'?admin_action=logout">'.c_comdef_htmlspecialchars($localized_strings['comdef_server_admin_strings']['logout']). ($user_obj->GetLocalName() != '' ? ' ('.$user_obj->GetLocalName().')</a></h4>' : '</a></h4>');
```
jbraswell: Can you change this so that `</a></h4>` is appended at the end instead of being repeated?

Issue #43, opened 2018-11-12
Add setting for county dropdown menu, similar to $meeting_states_and_provinces


jbraswell:

pjaudiomv:
if this does get implemented it obvs shouldnt overwrite existing data and only geocode if empty

pjaudiomv:
this is done and will be in next release

jbraswell:
Should open another issue for that.

Issue #44, opened 2018-11-12
Automatically fill in the county field when a meeting is saved


jbraswell:

jbraswell:
Could be in conflict with #43

pjaudiomv:
if this does get implemented it obvs shouldnt overwrite existing data and only geocode if empty

Redguy1959:
I would be satisfied with the County be a drop down box rather than a geocode, or at least have a drop down box override.

pjaudiomv:
when geocoding google will return suffix of County ex. "Los Angeles County" I would also like to see that suffix removed as County is implied because the field name is County. so return just this "Los Angeles". after looking through tomato for `?switcher=GetFieldValues&meeting_key=location_sub_province` a good majority doesn't have the County suffix either.

jbraswell:
I wonder what is returned for county for non-US locations. Do our non-US users fill out the county field at all?

jbraswell:
If we did this, we would be disabling manual editing of the County field, and relying completely on the geocoding API to fill it in. This is probably fine 99.9% of the time in the US. My fear is that someone outside of the US may be using the field for something that wouldn’t be returned in administrative_area_level_2. I guess we could go through all of the non-US root servers in the world in tomato and see if anyone uses the field.

pjaudiomv:
i know Ireland does

jbraswell:
Same concerns also apply to https://github.com/bmlt-enabled/BMLT-Root-Server/issues/45

Short-term solution for both would be to only do this for US locations.

pjaudiomv:
I don't know how much I like the idea of disabling manual entry of county or zip fields, I'd rather see just dropdown option #43 if thats the case. though that just my opinion. if there was somehow a localization option or something that could only enable this if option is set in autoconfig id feel alot better about it. but possibly degrading the system for others outside the us is not good

jbraswell:
Yeah it’s just not possible to do this without effectively disabling manual editing of the fields unless we modify the database schema to store whether or not things are overridden.

Only filling in the County or Zip if they were previously empty is an option. 

Making this an opt-in setting could also work.

pjaudiomv:
I like the idea of an opt in setting

jbraswell:
So I’m seeing three options, and they’re mutually exclusive 

1. Only do this for US addresses
2. Only set the County and Zip Code fields from geocoding results when they were previously empty. 
3. Require that users opt-in to this feature with a setting.

[...]

PR #121, opened 2018-12-20
Allow specification of county select menu items

pjaudiomv:
referencing issue/feature request #43


pjaudiomv:
I like this a lot better than #44  this would degrade performance for others, we can't reliable do #44 and results could be unpredictable for other countries. if #44 did get implemented you would have to still be able to manually edit as well. as @LittleGreenViper would say i don't think the juice is worth the squeeze.

pjaudiomv:
really travis? for linting too? me and you are gonna have a long talk one of these days.

pjaudiomv:
we could additionally choose to kill of unstable and bring it back to merge there. but ive tested this and feel pretty good about it. doesn't matter to me.

PR #125, opened 2018-12-24
composer update

pjaudiomv:


composer.json
```diff
@@ -3,7 +3,7 @@
         "vendor-dir": "main_server/vendor"
     },
     "require": {
-        "bmlt/bmlt-satellite-base-class": "^3.9"
```
jbraswell: Why was this necessary?

pjaudiomv: honestly maybe patience, I updated packagist but it wasn't upgrading to 3.9.11 once i set it to greater than 3.9.10 it worked. I can change it back. I bet if i had waited 20 minutes it would have worked

pjaudiomv: I've had this problem before when trying to update dependencies that they themselves just go tupdated only minutes ago.

jbraswell: Yeah I’m not positive but I don’t think this change is necessary and probably should be reverted. After that, ship it!

pjaudiomv: no it most def shouldn't be like I said Ive only ran into this issue when the dependency totes just got updated itself, we shall find out on next build.

pjaudiomv: I guess in retrospect if i did need to update lock file like that, I should of never pushed the commit change to the composer.json file. wasn't thinking

PR #128, opened 2018-12-27
add 'secret' option to not sort formats

pjaudiomv:
if `$sort_formats = false;` is set in auto-config the formats won't be sorted.


pjaudiomv:
I like that idea

radius314:
source

PR #130, opened 2019-01-14
adding meetings by location and day option to bmlt.js

pjaudiomv:
this would allow you to get meetings by location. the following needs to be sent to it
day, geoWidth, position
day = int 1-7
geoWidth = neg or pos int same as bmlt
position this would be long/lat object like so
myobject { latitude: 31.060, longitude: -82.154}

Example of this in action
https://bmlt.charlestonna.org/meetingsByDayLocation.html

and gist
https://gist.github.com/pjaudiomv/1db3c00ac6c81559de6b96940e69aba7

### DIFF
--- a/main_server/local_server/bmlt.js
+++ b/main_server/local_server/bmlt.js
@@ -89,6 +89,10 @@ var getServiceBodyIdQueryString = function (serviceBodyIds) {
     return serviceBodyIdString;
 };
 
+var getMeetingsByLocation = function (day, geoWidth, position, callback) {
+    getJSON(bmltbaseURL + "GetSearchResults&weekdays=" + day + "&geo_width=" + geoWidth + "&long_val=" + position.longitude + "&lat_val=" + position.latitude + "&callback=?", callback);
+};
+
 var getJSON = function (url, callback) {
     var random = Math.floor(Math.random() * 999999);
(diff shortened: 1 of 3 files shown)


main_server/local_server/bmlt.js
```diff
@@ -89,6 +89,10 @@ var getServiceBodyIdQueryString = function (serviceBodyIds) {
     return serviceBodyIdString;
 };
 
+var getMeetingsByLocation = function (day, geoWidth, position, callback) {
```
radius314: can you move this above line 64

pjaudiomv:
Ok not %100 sure I am I think I want to add sort_keys=start_time to it. It seems like tomatos default sorting is different than stock root servers

pjaudiomv:
yup tomato not sorting properly, however I dont feel like I shoould account for that here rather fix it in tomato. Im gonna adjusrt version then squash

jbraswell:
Please describe the sorting behavior you see and what you expect. If you
aren’t specifying a sort order it’s probably just returning them in the
order they’re stored on disk in the database.

pjaudiomv:
yea just submitted issue on tomato https://github.com/bmlt-enabled/tomato/issues/40 looking into it more. I believe thats what its doing just returning order they’re stored on disk in the database

Issue #134, opened 2019-01-16
Add the ability to override the 5 minute interval for start_time and duration.


radius314:
The default `start_time` and `duration` interval in the admin ui is hardcoded to `5`.  Would like the ability to override this to set it say `1` minute.

radius314:
working on this now.

radius314:
#135 is open

PR #135, opened 2019-01-16
Added $default_minute_interval to override the minute interval for Start Time and Interval on the Admin UI. #134

dgershman:


jbraswell:
If you’ve tested the above cases, feel free to merge, but please do a squash merge.

pjaudiomv:
I tested all three of those use cases as well

jbraswell:
Cool. Then we just need to resolve conflicts and then squash merge.

PR #144, opened 2019-04-20
Improve Install Wizard

jbraswell:
Changes:

1. Improved error display on the final page of the wizard. The error messages are now more clear, and there is now a link under each error taking the user to the to the correct tab to fix the problem. These are the possible error conditions:
   - Bad database information
   - Bad Google Maps API Key
   - Invalid server administrator username or password
2. Modified the wizard to automatically write `auto-config.inc.php`.


jbraswell:
1. Rename `Initialize Database` button to `Initialize Root Server`
2. Have the `Initialize Root Server` both initialize the database and attempt to write `auto-config.inc.php`.
3. If `auto-config.inc.php` cannot be written (likely due to permissions), print a warning and allow the user to write their own `auto-config.inc.php` file.

jbraswell:
Remaining work...

1. If the config file is written successfully, redirect the user to `/main_server/` so that they can log in to their new root server.
2. In the event that the config file cannot be written, show the config file in the UI and let the user know that they'll need to write the file themselves.
3. Remove unused strings from translation.

jbraswell:
From a technical standpoint, this is done, but I want to review and probably modify the text on each page of the install wizard before merging. Reminder: this includes the text that is displayed with the config file fails to write.

PR #159, opened 2019-04-28
Add ability to update World IDs from NAWS Spreadsheet through Server Administration UI

jbraswell:
1. Added "Server Administration" section to the server admin user interface. Only the server administrator user can see it.
2. The "Server Administration" section of the server admin user interface contains a form for uploading the NAWS spreadsheet that contains updated World IDs.
3. The server accepts both excel spreadsheets and CSV files. The parsing is handled by a new the phpspreadsheet library, a composer dependency.
4. The UI displays a report outlining number of meetings updated, not updated, and meetings that were not found in the database.
5. Because we added a new dependency, we locked down the vendor dir with a .htaccess file. Only static files can be retrieved from `vendor/`.


jbraswell:
- [ ] Ensure World IDs are being parsed out of the spreadsheet and saved correctly.
- [ ] Ensure the change history for each updated meeting is correct.
- [ ] Show the throbber animation in the user interface while the server is processing.
- [ ] Return a list of updated IDs from the server, and display them in the UI.
- [ ] Add lots of error handling.
- [ ] Add a `.htaccess` file to prevent direct access of everything in the `vendor` directory except for the bmlt satellite base class.

PR #163, opened 2019-05-13
Update csv.php

otrok7:
Added format_type_enum to output


jbraswell:
Thanks @otrok7! Will you add a comment with some context explaining why it's important to have `format_type_enum` in the output?

otrok7:
The format_type_enum allows the satellite to distinguish between different types of formats, so that they can be displayed differently.  For instance, languages can be displayed as flags.  Formats indicating special interests (mens/woman/LGBQ/Open) can be handled differently from formats describing the particulars of the meeting, etc.

jbraswell:
Cool. Does the user interface provide a way to change the `format_type_enum`?

pjaudiomv:
It currently does not and would have to be added as well.

pjaudiomv:
Would this field be populated with just integers or would this allow text as well

jbraswell:
It’s a `varchar(7)` field in the database, and has a default value of `FC1`.

pjaudiomv:
ah cool, the illusive FC1

otrok7:
I would like to add the UI as a separate pull request...  In the database it's a string, and I think that makes sense for the UI, too.  In the German speaking region we have 'LANG' for language, 'O' for open (German meetings are usually open once per month, so there are formats O1, O2, thru O5, that all have this format, plus O for always open).  Then we have the enums FC1 thru FC3, as described in Chris's code.  I don't know if it makes sense to standardize these.  Maybe different regions need different format types.  On the other hand, if we do standardize them, then we can allow satellites like the meeting map or bread to rely on them and use them.  This is pretty useful.

pjaudiomv:
ah cool found it
brief The format code type
        // - An enumerated string value, representing the classification of the format:
        // - "FC1" -Meeting Format (Speaker, Book Study, etc.)
        // - "FC2" -Location Code (Wheelchair Accessible, Limited Parking, etc.)
        // - "FC3" -Attendance Restriction (Men Only, Closed, Open, No Children, etc.)

jbraswell:
When new formats are created, what what `format_type_enum` do they get? Are they just `FC1`?

I am thinking that I am not comfortable merging this until there is a sane way to set the `format_type_enum` through the user interface.

pjaudiomv:
newly created formats set format_type_enum to null

pjaudiomv:
I agree though I would just leave this open and add commits for UI

otrok7:
I could add to the interface where formats are defined either a text-field, which would give the admin the ability to define his own types (tho its not really an enum, then) or a select box with options FC1, FC2, FC3 O and LANG.  What do you think?

jbraswell:
My preference is the select box, and for the time being not allow users to add additional enums.

I think it makes sense to translate `FC1`, `FC2`, and `FC3` to something human readable in the UI.

If you decide to take on this work, will you just do the work as additional commits in this pull request?

otrok7:
Hi, I've added the UI in my fork, but I don't seem to be able to create another pull request.  I did the commit from my IDE, and so it took my windows login as the author of the commit, is that the problem?

otrok7:
Basically, all I did was cut and paste the code for the format's NAWS Id

jbraswell:
The commits ended up in this pull request, so that's perfect.

It looks like the linting stage in the CI failed.

otrok7:
I see where it says there a 5 problems, but I can't find the description of where the problems are...

pjaudiomv:
we lint to the PSR-2 coding style, you could run the csv.php file through phpcbf to fix automatically https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically

pjaudiomv:
you can just push your commits to your fork and they will reflect automatically on this PR no need to open new one.

pjaudiomv:
you can either install phpcs and phpcbf locally with composer and add to your env to be available globally or download the phar and run from there. depending on environment it'd be something like this.
curl -sSL https://squizlabs.github.io/PHP_CodeSniffer/phpcbf.phar -o phpcbf.phar
php phpcbf.phar csv.php or whatever

pjaudiomv:
so since the linting failed on just that file you could run this as example`phpcbf --warning-severity=6 --standard=PSR2 --extensions=php /path/to/c_comdef_admin_main_console.class.php`

otrok7:
All checks have passed...

pjaudiomv:
Can you add the server admin strings to the rest of the languages as well, this can be just in English with maybe comments  // Todo translate or something  at end of each line

pjaudiomv:
Also bump version in serverInfo.xml to 2.13.1 and update Change list in the readme to reflect changes.

otrok7:
Done

pjaudiomv:
looks good, two things i would change. if no value for format_type_enum exists or is null it should still return the key format_type_enum with just empty value in response. this is to remain consistent with how the rest of keys and other semantic endpoints operate. second i think the additional format_type_enum values should also stay consistent and just be FC4,FC5 instead of O and LANG. would be curious what others think @jbraswell @BMLTGuy @RiftValleySoftware @radius314

RiftValleySoftware:
Just FYI. If we will be adding new fields to the semantic output, we need to make sure that they are available in all three outputs: CSV, XML, and JSON, and that the schema reflects the new field.

I know that XML is for squares, but if we add schema validation to the CI, then that actually tests a fairly wide swath of functionality.


THE GREAT RIFT VALLEY SOFTWARE COMPANY
	
CHRIS MARSHALL
President and Chief Executive Officer
Telephone: +1-631-406-3579 (Cell, Text)
Email: chris@riftvalleysoftware.com
1 Hewitt Square #308, East Northport, NY 11731 USA
https://RiftValleySoftware.com

pjaudiomv:
ah yes the XSD should be updated and good idea.

otrok7:
XML and JSON both work without any further modifications.

otrok7:
If you want me to change the codes, then it's FC1, FC2, and FC3 that should be changed, since the point would be to make the codes more mnemonic.   But I don't think it matters, any more than the numbers used for the formats do, because they are not visible to users.   I'd rather make the languages configurable, that's a separate pull request (I think).

RiftValleySoftware:
Does XML validate?

You should be able to simply feed the XML into a standard validator.

THE GREAT RIFT VALLEY SOFTWARE COMPANY
 <https://riftvalleysoftware.com/>	
CHRIS MARSHALL
President and Chief Executive Officer
Telephone: +1-631-406-3579 <tel:+1-631-406-3579> (Cell, Text)
Email: chris@riftvalleysoftware.com <mailto:chris@riftvalleysoftware.com>
1 Hewitt Square #308, East Northport, NY 11731 USA
https://RiftValleySoftware.com <https://riftvalleysoftware.com/>

pjaudiomv:
the xsd in these files should probably be updated `GetFormats.php` and `GetSearchResults.php`

otrok7:
Schema is updated

RiftValleySoftware:
I like Ron's style. Extremely disciplined. Can tell this ain't his first rodeo.

main_server/local_server/server_admin/server_admin_javascript.js
```diff
@@ -4268,7 +4268,7 @@ function BMLT_Server_Admin()
         in_offset = parseInt(in_offset, 10); // Just to make sure.
         var format_line_tr = null;
 
-        var insertion_point = (g_formats_array.length + in_offset) * (g_langs.length + 1);
+        var insertion_point = (g_formats_array.length + in_offset) * (g_langs.length + 2);
```
jbraswell: Will you explain the change on this line? I'm sure it's valid, I just want to understand.

main_server/local_server/server_admin/server_admin_javascript.js
```diff
@@ -4500,6 +4500,9 @@ function BMLT_Server_Admin()
         };
             
         container_row = in_container_table.insertRow(insertion_point);
+        if ( insertion_point > 0 ) {
```
jbraswell: Will you explain this change? With a comment, perhaps.

main_server/local_server/server_admin/lang/es/server_admin_strings.inc.php
```diff
@@ -42,169 +42,169 @@
...
-                                            'account_email_label'                                   =>  'Mi dirección de correo electrónico:',
-                                            'email_address_default_text'                            =>  'Introduzca una dirección de correo',
-                                            'account_description_label'                             =>  'Mi descripción:',
-                                            'account_description_default_text'                      =>  'Escriba una descripción',
-                                            'account_change_button_text'                            =>  'Cambiar configuración de mi cuenta',
-                                            'account_change_fader_success_text'                     =>  'La información de la cuenta se cambió con éxito',
-                                            'account_change_fader_failure_text'                     =>  'La información de la cuenta no se ha modificado',
+                                            'change_password_label'                                 =>  'Cambiar mi contraseÃ±a para:',
```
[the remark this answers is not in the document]
jbraswell: Yeah... this will need to be reverted

jbraswell:
I noticed that `initialFormatsData.sql` sets the "Open" and "Closed" formats as type `FC3`, which is the special interest/common needs format type. If we are going to add an additional format type for Open/Closed (we should, it's a good idea), then we should fix the `INSERTS` that file.

There's currently no good way to add that kind of database migration. We currently handle that sort of thing on every single request in `db_connect.php`. For now I'd probably just go ahead and add an additional try/catch block in there that performs an `UPDATE`. In the future we need to add actual database migration versioning so that we aren't running all this DDL on every request.

main_server/local_server/server_admin/lang/de/server_admin_strings.inc.php
```diff
@@ -344,6 +344,16 @@
                                                                                                             'Y'     =>  'Junge Menschen',
                                                                                                             'LANG'  =>  'Fremdsprache'
                                                                                                             ),
+        'format_type_prompt'                             =>  'Format Type:',
+        'format_type_codes'                                    =>  array (
+            ''      =>  'None',
+            'FC1'  =>  'Meeting Format (Speaker, Book Study, etc.)',
+            'FC2'  =>  'Location Code (Wheelchair Accessible, Limited Parking, etc.)',
+            'FC3'  =>  'Special Interest and Restrictions (Mens Meeting, LGTBQ, No Children, etc.)',
```
jbraswell: I think we should call this `Common Needs and Restrictions` instead.

jbraswell:
Moved to https://github.com/bmlt-enabled/bmlt-root-server/pull/173

Issue #169, opened 2019-05-22
Make semantic administration a default setting


radius314:
Seems pointless _now_ to make this an optional setting.  I recommend make this default behavior.

pjaudiomv:
I agree

RiftValleySoftware:
Agreed

RiftValleySoftware:
A lot of the defaults were set the way they were, because tinfoil.

pjaudiomv:
this is in master and will be in next release 2.13.1

PR #170, opened 2019-05-27
Make semantic administration default #169

pjaudiomv:
#169

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ CHANGELIST
 ----------
 ***Version 2.13.1* ** *- UNRELEASED*
 
+- Semantic Administration is now enabled by default on install wizard.
 - Fixed a bug with JSONP endpoint emoting incorrect mime type.
 
 ***Version 2.13.0* ** *- April 28, 2019*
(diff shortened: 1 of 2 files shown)


pjaudiomv:
I tested this and works fine

jbraswell:
When you tested this did it work fine?

pjaudiomv:
yea dude how many time do you need me to say it lol

jbraswell:
You don’t need to set `checked="checked"`. You can just have the `checked` attribute without setting it to anything. See https://www.w3schools.com/tags/att_input_checked.asp. 

I think removing the setting altogether might be better, though. I don’t think there’s much point in even having the option to disable the semantic UI. Thoughts?

pjaudiomv:
yes i know this, but thats how Meetings Are Considerd "CLOSED" by Default setting did it so i was just staying consistent. that must of been older code because there are other places where it is just checked

pjaudiomv:
note that this doesn't disable the semantic ui it disables the semantic ADMIN ui.

jbraswell:
That’s what I said bro look at my comment

jbraswell:
I’m totes fine with this getting merged. Up to you if you wanna do more here.

pjaudiomv:
yea i kind of just assume leave it there may be other tinfoil hats that want the option to disable.

jbraswell:
Works for me

Issue #47, opened 2018-11-13
Don't auto close meeting after save


pjaudiomv:
when editing a meeting after hitting save, the meeting is automatically closed and gos to just list view. this can be annoying when editing multiple meetings as you can get lost.

PR #189, opened 2019-07-06
Don't close meeting editor after save

jbraswell:
Closes https://github.com/bmlt-enabled/bmlt-root-server/issues/47

### DIFF
--- a/main_server/local_server/server_admin/server_admin_javascript.js
+++ b/main_server/local_server/server_admin/server_admin_javascript.js
@@ -1549,7 +1549,7 @@ function BMLT_Server_Admin()
                 eval('var json_object = ' + in_http_request.responseText + ';');
             };
         };
-            
+
         if ( json_object ) {
             var meeting_changed = false;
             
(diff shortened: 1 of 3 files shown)


main_server/local_server/server_admin/server_admin_javascript.js
```diff
@@ -1587,6 +1589,9 @@ function BMLT_Server_Admin()
             
             this.cancelMeetingEdit(in_orig_meeting_id, true);
             this.createMeetingList();
+            if (json_object && json_object[0] && json_object[0].id_bigint) {
```
jbraswell: What this actually does is re-open the same meeting.

pjaudiomv:
This may be my favorite change yet.

main_server/local_server/server_admin/server_admin_javascript.js
```diff
@@ -1569,6 +1569,8 @@ function BMLT_Server_Admin()
                                 this.createOneMeetingNode(single_meeting_div, this.m_search_results[c]);
                                 meeting_changed = true;
                             };
+
+                            break;
```
jbraswell: There's no reason to keep iterating after the meeting is found, so just `break`.

Issue #82, opened 2018-12-02
Unpublished notes / reason / the "why"


dgershman:
When a meeting is unpublished, it's for a variety of reasons.  Being able to document the "why" in a notes field of some sort and then have that exported to NAWS will help greatly.

radius314:
This would ideally have some canned responses.

pjaudiomv:
Yes, maybe even force to make a selection or type something. I administer two servers and often get people who should be deleting meetings rather than unpublishing. However I can't always be sure so this would be super helpful in general not just to naws. Obviously some of that can be helped by better training on using the server.

ChrisMarshallNY:
I just had a bit of back-and-forth with Danny on this. I think I understand what we need, and have a couple of ideas.

The main deal is to try to avoid a DB schema change.

With that in mind, we could repurpose the `published` column. It is currently a `tinyint(4)`, which carries either `1` (published) or `0` (unpublished -default).

We could create an enum, based on int values, which map to concrete classes of reasons, like so:

`0 - unpublished (no reason given)`
`1 - published`
`2 - unpublished - moving venue`
`3 - unpublished - seasonal`
`4 - unpublished - Crazy Joe threw a tantrum in the parking lot, and the meeting is taking some "down time."`

etc...

ChrisMarshallNY:
We might also be able to use the `change_name_string`. It is a rather tenuous field (and it also has a `description`). I suspect I don't use them much. These are the change record; not the meeting data.

pjaudiomv:
hidden could be better term than unpublish ? prob could use better UI indicator for unpublished (right now meetings just turn yellow)

PR #190, opened 2019-07-06
WIP: Add unpublished reason

jbraswell:
- [x] Add unpublished reason to UI
- [x] Handle unpublished reason on the backend
- [ ] Test iOS admin app
- [ ] Determine appropriate canned responses for unpublished reason
- [ ] Add `unpublished_reason` column to the NAWS csv
- [ ] Add message to the UI explaining that unpublished should not be used to indicate a meeting has permanently closed.
- [ ] Add strings to translation files
- [ ] Make sure cache busting is added to the js file before merging

Closes https://github.com/bmlt-enabled/bmlt-root-server/issues/82

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ CHANGELIST
 ***Version 2.13.3* ** *- TBD*
 
 - When saving a change to a meeting, the edit screen for that meeting is no longer closed. This makes it easier for the user to keep their place in the user interface.
+- Added an 'Unpublished Reason' field to the meeting editor UI to allow users to communicate why a meeting is unpublished. This will help NAWS reconcile BMLT lists with their database.
 
 ***Version 2.13.2* ** *- July 4, 2019*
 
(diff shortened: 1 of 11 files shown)


pjaudiomv:
Loss of meeting facility or temporary loss

jbraswell:
@pjaudiomv I think the core functionality here is mostly done. Could us some help with testing. I think I've tested most of it myself, but one gap I know about is the iOS app.

pjaudiomv:
if a meeting is already unpublished in the database the `"Save the Changes to This Meeting"` button is white implying theres been a change, if you then try to go to another meeting it asks you `"You are currently editing another meeting. Do you want to lose all changes in that meeting?"`even though I didnt make any changes. If I do select an `Unpublished Reason` and hit save it still does this, even if I go to another meeting and come back.

pjaudiomv:
another weird thing. the email_contact field in the database was always NULL but any one that i've edited turns to just empty even though i never edited that field.

Screen Shot 2019-07-07 at 7 47 04 PM

ChrisMarshallNY:
The "dirty flag" issue is familiar.
The JS does a pretty "brute force" test to see if a meeting  has changed; just a simple string test.
If there is ANY difference between the current meeting string, and the initial JSON object, then it flags it as changed. This includes things like the order of data members in the serialized object.

PR #195, opened 2019-07-23
make meetings default to published when creating a new meeting

### LINKED ISSUE #194: better checks, clearer ui for publishing meeting
lots of users create meetings and forget to check the `meeting is published` checkbox, this should be addressed with some sort of warning or something. maybe change default. I've seen this mistake a lot and however we make this clearer I'm fine with.

pjaudiomv:
Per discussion from last scrum. Many service bodies have had problems with users creating meetings and not publishing them, this is more than just not training them properly. Meetings will now be published by default when creating a new meeting. If a root server wants to reverse this change they can add `$default_meeting_published = false;` to their `auto-config.inc.php`

resolves #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,11 @@ For instructions on installing the root server, see [the page on installing a ne
 
 CHANGELIST
 ----------
+***Version 2.13.4* ** *- UNRELEASED*
+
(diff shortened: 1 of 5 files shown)


jbraswell:
If you’ve tested this, I’m good with it.

pjaudiomv:
Cool yup all tested.

PR #196, opened 2019-07-24
adding server version to login page

pjaudiomv:
@radius314  had suggested having server version on login screen, I often had thought/wished it would be there too. now it can be....  this would appear the same as when authenticated only white

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ CHANGELIST
 
 - Meetings are now published by default when creating a new meeting. To change this back to default to unpublished you would need to add the following to your `auto-config.inc.php`.
     - `$default_meeting_published = false;`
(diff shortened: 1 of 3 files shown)


pjaudiomv:
could go bottom right too idk

pjaudiomv:
Screen Shot 2019-07-23 at 11 32 20 PM
Screen Shot 2019-07-23 at 11 29 44 PM

pjaudiomv:
Bottoms better

radius314:
How’s it look on mobile?

radius314:
It could also be centered under the log in button.

pjaudiomv:
it actually looks better on top right for mobile

PR #199, opened 2019-08-06
Tweaked the Doxygen file to generate valid docs.

ChrisMarshallNY:
Added the auto-generated docs directory (for GH Pages).

I added a "docs" directory, so we can have a GH Pages site for the Root Server.

Currently, the docs directory is created and populated by Doxygen (using doc/Doxyfile).


pjaudiomv:
I am totes cool with this, however I think it should be auto generated and deployed upon commit to master. this could be better off in its own branch with just the docs to make automating easier however may not be needed. travis would commit and deploy the docs to whatever branch is decided upon.

ChrisMarshallNY:
Well, the whole deal is to associate the docs directory with the repo, so you can have a GitHub Pages for the repo. It needs to be part of the repo. Since the distro is actually a subdirectory of the repo, the docs don't get distributed.

Doxygen is a CLI tool that can easily be run by Travis.

That said, we could do the docs differently. Doxygen is a fairly "blunt instrument."

pjaudiomv:
yea i guess there is no reason to have separate branch, however if it did make sense, you could still generate them from one branch and push the docs to another. just found gist with example of using travis with doxygen https://gist.github.com/vidavidorra/548ffbcdae99d752da02

pjaudiomv:
look like travis has pages as a provider https://docs.travis-ci.com/user/deployment/pages/

ChrisMarshallNY:
If you guys want to set up a docs repo, you can. I find it a bit better to keep as close as possible to "ground truth." The docs are generated by comments in this repo, alone, so I'd be most comfortable having the doc generation and storage attached to it directly.

pjaudiomv:
I totally agree if its docs for the root server, it should def be in the root server repo.

pjaudiomv:
i'm going to test auto generating on a separate branch.

PR #214, opened 2019-09-20
Place name_strings from unmapped formats in the Room field of a NAWS Export 

### LINKED ISSUE #209: Place name_strings from unmapped formats in the Room field of a NAWS Export
Requested by Lori to get more accurate meeting information from formats that don't necessarily map to NAWS formats.

pjaudiomv:
Resolves #209


pjaudiomv:
something funky is happening with csv escaping here. when thats solved it should be fine.

PR #215, opened 2019-09-20
WIP: adding new NAWS formats

### LINKED ISSUE #212: Guiding Principles as NAWS mapping option
Can we get Guiding Principles (GP) as an option to map a Guiding Principles meeting?
If this is the wrong place to suggest ask this... just point me in a different direction.

Thanks :)

pjaudiomv:
closes #212


pjaudiomv:
I actually think this is pretty good, only question is here. https://github.com/bmlt-enabled/bmlt-root-server/pull/215/files#diff-b12210ac32631adc90f6385ac673b212R56

PR #216, opened 2019-09-21
fix for naws dropdown sorting by key instead of value

pjaudiomv:
the last fix was silly, this one may be too. But i was sorting by key and really needed to sort by value. Because the default dropdown is the top one and that key is null. im shifting it off sorting and then unshifting it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ CHANGELIST
 - Changed JSONP mime type to `application/javascript`
 - Rename "Get A NAWS Format Dump" to "Get A NAWS Format Export" in the semantic workshop
 - Fixed an issue where the admin user interface could falsely think a meeting had been edited
-- Fix for NAWS format drop down not sorting correctly.
+- Fix for NAWS format drop-down not sorting correctly in the root server administration.
 
 ***Version 2.13.5* ** *- August 15, 2019*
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -63,6 +63,9 @@ CHANGELIST
 - Rename "Get A NAWS Format Dump" to "Get A NAWS Format Export" in the semantic workshop
 - Fixed an issue where the admin user interface could falsely think a meeting had been edited
 
+
```
jbraswell: I think you've got an extra line here

main_server/local_server/server_admin/c_comdef_admin_main_console.class.php
```diff
@@ -381,7 +381,9 @@ public function return_main_console_html()
                 $ret .= 'var g_naws_popup_prompt = \''.self::js_html($this->my_localized_strings['comdef_server_admin_strings']['world_format_codes_prompt']).'\';'.(defined('__DEBUG_MODE__') ? "\n" : '');
                 $ret .= "var g_naws_values = [";
                     $n_first = true;
-            ksort($this->my_localized_strings['comdef_server_admin_strings']['world_format_codes']);
```
jbraswell: This looks good, but it'd be helpful to put a comment in the code explaining why you're doing what you're doing

pjaudiomv: ok I added notes.

README.md
```diff
@@ -62,6 +62,7 @@ CHANGELIST
 - Changed JSONP mime type to `application/javascript`
 - Rename "Get A NAWS Format Dump" to "Get A NAWS Format Export" in the semantic workshop
 - Fixed an issue where the admin user interface could falsely think a meeting had been edited
+- Fix for NAWS format drop down not sorting correctly.
```
jbraswell: “Fixed an issue where <describe the issue>”

PR #219, opened 2019-09-29
WIP: Allow import of NAWS export in installation wizard

jbraswell:


main_server/local_server/install_wizard/installer.js
```diff
@@ -737,10 +747,10 @@ function gmScriptLoadCompletion()
 *   \returns a new XMLHTTPRequest object.                                                   *
 ********************************************************************************************/
     
-function BMLT_Installer_AjaxRequest(
+function BMLT_Installer_AjaxPostRequest(
```
jbraswell: All usages of this function were for `POST`, so I simplified it by removing the `method` option.

pjaudiomv:
None of your commits are verified, Please sign your commits and re open.

main_server/local_server/install_wizard/installer_ajax.php
```diff
@@ -30,6 +30,11 @@
 require_once(dirname(__FILE__).'/../../server/shared/classes/comdef_utilityclasses.inc.php');
 require_once(dirname(__FILE__).'/../../server/shared/Array2Json.php');
 
+function generateRandomString($length = 10)
```
jbraswell: Used to generate random passwords. This is not really a cryptographically secure random string generator, but it's close enough for BMLT purposes.

jbraswell:
- [ ] Write meeting creation code
- [ ] Remove all TODOs
- [ ] Front-end logic for handling situation where config couldn't be written
- [ ] Front-end logic for displaying import errors
- [ ] Figure out what to do if an import fails - should it delete the config file and de-initialize the db? Should it just delete the imported data?

PR #221, opened 2019-10-01
skip docs if not master

radius314:


pjaudiomv:
whats the difference between the two

jbraswell:
I am not convinced that we don't want to build the docs on non-master. Seems useful to know if the build is broken. It doesn't deploy them.

pjaudiomv:
how is `on branch: master` different than `if: branch = master`

radius314:
Is the placement of the conditional.  One is on the job lifecycle the other is on the entire stage.

pjaudiomv:
oh shoot yea thanks, probably would of figured out looking at whole file

PR #230, opened 2019-10-26
Farsi formats

otrok7:
These are the formats as translated by Meysam


pjaudiomv:
not sure if its a encoding issue or what, but setting up a new server using the wizard is failing with this. It creates all the tables however the `na_comdef_formats` table is empty and doesn't get populated with `InitialFormatsData.sql`  this is the error I get on wizard.

Screen Shot 2019-10-27 at 9 06 01 AM

pjaudiomv:
so found problem with the sql the last line should end with a `;` not a `,` comma

pjaudiomv:
after I changed that it imported fine

pjaudiomv:
next thing is for the formats to show up in the `Format Administration` of server admin their must be a folder in `/main_server/local_server/server_admin/lang` named `fa`

pjaudiomv:
this could probably just be a copy of the english `en` folder with the `name.txt` file renamed to farsi

pjaudiomv:
ok just fixed sql and pushed the lang folder, will want to test this build when it finishes.

pjaudiomv:
this build contains those changes https://archives.bmlt.app/bmlt-root-server/bmlt-root-server-build881-03b78de938359374d2db0baa31d09cc2fb2d85fa.zip

otrok7:
I added a feature for format languages <> root server languages a while ago, to support exactly this usecase.  The point is to focus for now on the feature that serve Farsi speaking members outside Iran....at least until there's a decision from the Iran Region(s).....But "faking it' with a copy of the English lang directory is just as good.

pjaudiomv:
cool yea latest build worked great

Screen Shot 2019-10-27 at 9 48 29 AM

PR #234, opened 2019-11-03
Show more useful api key error

jbraswell:
image


jbraswell:
We need to test this in a few different browsers...

- [ ] firefox
- [ ] chrome
- [ ] modern version of IE
- [ ] safari on macos
- [ ] safari on ios

Issue #224, opened 2019-10-06
Override config option for auto geocoding of addresses


paulnagle:
Would it be possible to have a config option to disable auto geocoding? Geocoding doesn't work so well in some areas of the world.

jbraswell:
@paulnagle, this is implemented in 2.14.6. If you set `$auto_geocoding_enabled = false;` in `auto-config.inc.php`, it will no longer automatically geocode meetings and the latitude/longitude fields become editable.

PR #252, opened 2019-12-08
Make automatic geocoding optional

jbraswell:
Closes https://github.com/bmlt-enabled/bmlt-root-server/issues/224

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ For instructions on installing the root server, see [the page on installing a ne
 
 CHANGELIST
 ----------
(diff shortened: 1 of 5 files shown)


pjaudiomv:
one thing youll have to do with this is then make the long/lat fields editable if auto geocoding is set to false

jbraswell:
Good point. I'll work on that.

jbraswell:
Okay, the lat/long fields are now editable when auto geocoding is disabled.

Issue #45, opened 2018-11-12
Auto fill zip from geocode results.


radius314:

jbraswell:
Now that auto-geocoding is configurable, it might make sense to do this.

jbraswell:
Closed by https://github.com/bmlt-enabled/bmlt-root-server/pull/255

PR #255, opened 2019-12-09
Automatically populate zip and county from geocode results

jbraswell:
Three settings:
- `$auto_geocoding_enabled`
- `$zip_auto_geocoding_enabled`
- `$county_auto_geocoding_enabled`

**`$auto_geocoding_enabled`**: The default value is `true`. Enabling `$auto_geocoding_enabled` by itself just turns on auto geocoding for the latitude and longitude fields.

**`$zip_auto_geocoding_enabled`**: The default value is `false`. If both `$auto_geocoding_enabled` and `$zip_auto_geocoding_enabled` are `true`, the zip code field becomes readonly, and is automatically populated by geocoding on save.

**`$county_auto_geocoding_enabled`**: The default value is `false`. If both `$auto_geocoding_enabled` and `$county_auto_geocoding_enabled` are `true`, the county field becomes readonly, and is automatically populated by geocoding on save.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ CHANGELIST
 ***Version 2.14.6* ** *- TBD*
 
 - Geocoding does not work well in all parts of the world, so automatic geocoding of meetings is now optional. It is enabled by default, and can disabled by adding `$auto_geocoding_enabled = false;` to `auto-config.inc.php`.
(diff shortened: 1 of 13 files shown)


jbraswell:
Closes https://github.com/bmlt-enabled/bmlt-root-server/issues/45

Closes https://github.com/bmlt-enabled/bmlt-root-server/issues/44

PR #285, opened 2020-03-27
add vm and tc formats

pjaudiomv:
not sure what to set `shared_id_bigint` to


pjaudiomv:
Ok I feel good about this, I have tested the following ways.

1. With a new root server
2. with an existing root server
3. with an existing root server that already has a vm or tc format.

feel free to make any changes though.

PR #286, opened 2020-03-27
adding vm and phon num fields to naws export

pjaudiomv:


pjaudiomv:
I ran phpcbf on the file.... It's mo beutiful

pjaudiomv:
actually this is redundant, theres no need for seperate function. as those fields will be there

PR #291, opened 2020-03-28
removing web format on new installs in favor of vm

pjaudiomv:
this will remove some pain in future possibly.  I didn't know this format existed so i had people in new england selecting this format instead of vm. for nerna I just ended up renaming the web format to vm and removing the newly created one.


jbraswell:
I am definitely fine with removing the format for new installs - I think that is all this PR does. 

I’d be interesting in finding out how many current root servers are using the WEB format.

pjaudiomv:
```
Arizona Region
Autonomy Zone
Canadian Assembly
Canadian Assembly
Connecticut Region
Greater New York Server
NA Colorado
NA Denmark
NA Kentucky
NA Sweden
Sao Paulo ASC
Show-Me Region
```

Issue #301, opened 2020-05-22
Add formats_comparison_operator


klgrimley:
Would be nice to use additional comparisons other then AND for format searches like you can with tomato. This came up on my Area site when needing to produce a listing of meetings that have either the HY OR VM format.

pjaudiomv:
This would be a nice feature as I have needed it before as well.

tempsaint:
Agreed...

jbraswell:
I think we should support something like this for pretty much all object types, not just formats

jbraswell:
This is in 2.15.3

tempsaint:
This worked excellently. Thank you. I was able to use formats_comparison_operator=OR to pull a custom query for two different formats. Nice job team.

PR #304, opened 2020-06-21
Add formats_comparison_operator to GetSearchResults for included form…

jbraswell:
…at filters


jbraswell:
Implements https://github.com/bmlt-enabled/bmlt-root-server/issues/301

Issue #306, opened 2020-06-21
Add formats_comparison_operator to semantic wizard


jbraswell:

radius314:
Was this delivered already, or is this something else?  I saw something on a semantic wizard install running 2.15.3.

jbraswell:
Yes

PR #308, opened 2020-06-27
Add formats_comparison_operator to semantic workshop

jbraswell:


jbraswell:
`formats_comparison_operator=OR` should only be added to the query string when formats are selected to be included and the `OR` radio button is selected

jbraswell:
Closes https://github.com/bmlt-enabled/bmlt-root-server/issues/306

PR #309, opened 2020-07-10
add virtual meeting additional info field

alanborning:
This change adds a new field "Virtual Meeting Additional Info" and a prompt for this field in the meeting editor. Virtual meeting listings often include additional information beside the URL. For example, zoom meetings may list a Meeting ID and a password, which allows people to join the meeting directly from zoom by entering the Meeting ID rather than by clicking the link, or to join the meeting from a landline phone by calling the access number and entering the Meeting ID. This information is currently being stored in various ad hoc fields; this consolidates it into a named place.


README.md
```diff
@@ -57,6 +57,9 @@ For instructions on installing the root server, see [the page on installing a ne
 
 CHANGELIST
 ----------
+***Version 2.15.4* ** *- TBD*
+- Added the new "Virtual Meeting Additional Info" field to the NAWS export.
```
jbraswell, line 61: I don't think this message is quite right. What you've done is added the field to the root server itself and to the semantic api output, but you haven't yet added the field to the NAWS export. I'd fix this message, go ahead and merge this, and then add the new field to the the NAWS export. You can see an example of how to do that here: https://github.com/bmlt-enabled/bmlt-root-server/pull/286.

alanborning: True - I fixed the README. I'll work on the NAWS export next. So the plan is to hold off on releasing this as version 2.15.4 until the NAWS export is fixed also, is that right?

jbraswell: I think we'll hold off on the next version until the NAWS export is fixed, yes, but we can go ahead and merge this PR. I'll let you click the merge button (I prefer squash merges, please). The idea is that `master` is always releasable and never broken, but we can still wait to cut a release until we think "enough" stuff is ready.

PR #310, opened 2020-07-11
passing gkey in as env var

pjaudiomv:


jbraswell:
I think if we merged this right now it'd make the CD hang, you'll need to find a way to pass in the variable to the terraform on each run before we merge this.

pjaudiomv:
I added the env var to travis so it should be available as soon as container spins up, `TF_VAR_GOOGLE_API_KEY`

jbraswell:
Will you add the variable to a `variables.tf` file instead?

pjaudiomv:
done

jbraswell:
If you say something nice about me, I'll merge this

pjaudiomv:
I miss the way you smell, which is good.

PR #313, opened 2020-07-12
add virtual_meeting_additional_info field to NAWS export

alanborning:
This adds NAWS export support for the virtual_meeting_additional_info field that was added in #309.


alanborning:
This adds NAWS export support for the virtual_meeting_additional_info field that was added in #309.

PR #314, opened 2020-07-12
debug capable docker image

radius314:


docker/Dockerfile-bmlt
```diff
@@ -13,4 +15,6 @@ RUN chmod +x /tmp/start-bmlt.sh
 RUN ln -sf /proc/self/fd/1 /var/log/apache2/access.log && \
     ln -sf /proc/self/fd/1 /var/log/apache2/error.log
 
+RUN echo "nameserver 8.8.8.8" >> /etc/resolv.conf
```
jbraswell, line 18: Why is hard coding the name server required? When running in AWS, I think I'd want it to default to the internal name servers of the vpc it's deployed to.

radius314: I don't remember the exact reason, but I hit issues with this a long time ago.  I believe related to the Google Maps API lookups during install, but it could be something else.  

How would we set it so that it would work both in AWS and outside AWS?

radius314: Also the existing implementation does this, I'm not adding any functionality here with that.  I think changing it out would require additional testing.

radius314: https://github.com/bmlt-enabled/bmlt-root-server/blob/master/docker/Dockerfile-apache2#L6

jbraswell: If the existing implementation does it, I'm fine with keeping it. That said, best practice is generally allowing DHCP of any given container orchestrator or local environment to provide DNS server. That _should_ work everywhere. If it doesn't, there's a problem with the image. But, again, if the existing image already does this, it's fine.

PR #315, opened 2020-07-12
fixing ulimit issue

radius314:
Attempting to address this issue in the failing ECS container.

image

### DIFF
--- a/terraform/ecs-service-bmlt-latest.tf
+++ b/terraform/ecs-service-bmlt-latest.tf
@@ -21,7 +21,11 @@ resource "aws_ecs_task_definition" "bmlt_latest" {
     "essential": true,
     "entryPoint": null,
     "mountPoints": [],
-    "ulimits": null,
+    "ulimits": {
+      "softLimit": 2048,
+      "hardLimit": 2048,
+      "name": "nofile"
+    },
     "dockerSecurityOptions": null,
     "environment": [
       {
--- a/terraform/ecs-service-bmlt-unstable.tf
+++ b/terraform/ecs-service-bmlt-unstable.tf
@@ -21,7 +21,11 @@ resource "aws_ecs_task_definition" "bmlt_unstable" {
     "essential": true,
     "entryPoint": null,
     "mountPoints": [],
-    "ulimits": null,
+    "ulimits": {
+      "softLimit": 2048,
+      "hardLimit": 2048,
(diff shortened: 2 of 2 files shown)


jbraswell:
What is the ulimit issue? How does it present itself, what is the impact?

radius314:
The container died.  I looked at the Cloudwatch logs and saw this. 

https://github.com/bmlt-enabled/bmlt-root-server/pull/315#issue-447948535

radius314:
I am referencing this as my attempt to address this https://stackoverflow.com/questions/57486473/ulimit-error-setting-limit-operation-not-permitted-using-apachectl-in-cont.

jbraswell:
In the logs, it looks like Apache is trying and failing to set the ulimit. Does having the ECS agent set it per the proposed task definition change prevent Apache from making the calls to set the ulimit?

radius314:
My thought was that it was trying to set it to `null`.  I honestly have no idea.

jbraswell:
Okay. If it's tested, let's merge it. I'd like to a avoid merging things to master that we aren't sure if they work.

radius314:
How can I test this ahead of merging to master?  I see an unstable ECS configuration, do i merge to unstable to try things out there?

jbraswell:
You could  do that, or you could manually modify the task definition (terraform will revert it if it doesn't work), or you could run the terraform locally with the change to see.

jbraswell:
I am not sure that's even a fatal error. It looks like apache still started and serves traffic just fine.

image

The real error is `/tmp/start-bmlt.sh: line 39: [: too many arguments`.

radius314:
I'm wondering if it's the default shell and if we should be explicit with a shebang.

jbraswell:
`if [ ! -z $DO_NOT_FORCE_PORT -a "$DO_NOT_FORCE_PORT" == "true" ]`

I wonder if `$DO_NOT_FORCE_PORT` needs to be quoted like this: `if [ ! -z "$DO_NOT_FORCE_PORT" -a "$DO_NOT_FORCE_PORT" == "true" ]`

radius314:
testing this theory now

jbraswell:
Pretty sure if you just surround $DO_NOT_FORCE_PORT in double quotes it'll fix it.

jbraswell:
Looks like it's been broken for a while?

PR #316, opened 2020-07-12
this will fix the one error pointed out

radius314:


radius314:
not sure if this will actually resolve the issue.

radius314:
i did test this and the error goes away for sure.

PR #320, opened 2020-07-13
dynamic theme

radius314:


pjaudiomv:
I would instead add a config option for `auto-config.inc.php` say `meeting_browser_theme` then in https://github.com/bmlt-enabled/bmlt-root-server/blob/master/main_server/server/c_comdef_server.class.php you could add

`c_comdef_server::$server_local_strings['meeting_browser_theme'] = isset($meeting_browser_theme) ? $meeting_browser_theme : (isset($_GET["theme"]) ? html_entity_decode($_GET["theme"]) : "florida-nights");` and reference this `theme: "<?php echo $console_object->my_server::GetLocalStrings()["meeting_browser_theme"] ?>",` for the theme in https://github.com/bmlt-enabled/bmlt-root-server/blob/master/main_server/client_interface/html/index.php

pjaudiomv:
this would allow people to set their own theme as well as use a get var. or default to florida

radius314:
go for it

radius314:
I would also eliminate the query string if you chose the config route

PR #328, opened 2020-07-19
WIP: Add HY format, map virtual formats to new NAWS format codes.

jbraswell:
This adds the Hybrid format to new and existing installations. For existing installations, a new `HY` format is only added if there is not already an `HY` format in the database. Any existing `HY`, `TC`, or `VM` formats are updated to map to the new NAWS formats.

This PR should not be merged until I receive confirmation of the new NAWS format codes from Lori.


pjaudiomv:
Can I merge this

alanborning:
How consistent do we want to be with the format codes? One standard would be to use two capital letters for two words (e.g., TC for Temporarily Closed), and a capital then lower case for the first two letters of the first word (e.g., Ti for Timer).  But this isn't done everywhere. So ... is VM and HY ok, or should it be either (Vi, Hy) or (VM, HM)?  Or just don't worry about this level of consistency? Hy or HY does seem more intuitive than HM.

jbraswell:
Good points re: consistency. I think we are stuck with `VM` and `HY`, because they are already in use by half the US.

We've just been winging it with formats so far, going with whatever feels right 🤷‍♂️.

alanborning:
I'm wondering about the different language versions. The PR has changes for most of the languages in files like this: `main_server/local_server/install_wizard/sql_files/InitialFormatsData-es.sql` (file for Spanish), but there isn't a file for German. (No file at all in the repo, not just in these changes).

And on the other hand there is a change in the PR just for German but not any of the other languages, although there are files for all the others: main_server/local_server/server_admin/lang/de/server_admin_strings.inc.php 

Other than my question about that, looks good!

jbraswell:
Right. I only modified what already exists in the repo. Sounds like no one ever added an initial formats sql script for populating the formats table with german translations on a fresh installation. Would be a good thing to add, but I think that would be best in another PR.

jbraswell:
@pjaudiomv you can merge this now

PR #330, opened 2020-07-27
install wizard improvements

### LINKED ISSUE #326: check for whitespace in Database Connection Settings
During setup, if you have a space after the Database User (or password, or database name), the database test will fail. As a minor improvement, if there is such whitespace I suggest adding a message noting that as a possible reason for the failure. (When I was first setting up a BMLT server, I futzed around for a little while trying to figure out what that problem was, and was just walking someone else through a BMLT setup last night and they had the same issue.)

If this seems like a reasonable improvement I will self assign it.

### LINKED ISSUE #327: setup wizard should escape single quotes in user names, passwords, and database names
The auto-config.inc.php file is malformed if there is a single quote in a user name, password, or database name -- user O'Toole is out.  To fix, the setup wizard should escape single quotes.

alanborning:
This PR is intended to close #326 and close #327.

For #326, in the installer if you test the database connection settings, it gives you a warning if one of your fields has whitespace at the beginning or end. My reasoning for making this a warning rather than simply an error is that in the use case I was considering (the user accidentally included a space at the end of a field, probably with cut-and-paste), the connection will also fail since the actual field doesn't have this space. So you do get an error, but also with a hint about the cause. However, if for some totally weird reason, you actually have say a password with a space at the end, you can use it; it will work, just give a warning if you test it. I tested this by trying spaces before and after each field. The warning is in the languages files (e.g., main_server/local_server/server_admin/lang/de/install_wizard_strings.php), but I didn't see a way to run the installer on anything but English, so only en/ is tested.

For #327 it escapes single quotes and backslashes when writing an auto-config.inc.php file.  So user O'Toole is now in business.


alanborning:
There will now be a couple of minor merge conflicts with PR #331 -- please review #331 first, and if that's accepted and merged, I can fix up this PR accordingly.

Issue #343, opened 2020-09-20
Root server should be time zone aware


jbraswell:
We've punted this for a long time, but now that virtual meetings are a thing and searches are no longer purely geographic, we need to do it. I propose the following.

1. Add a time_zone column next to the current meeting time column. It should be varchar.
2. Valid values for the column should be from https://www.iana.org/time-zones.
3. Only time zone names should be stored in the column. Not the offsets, because the offsets change. It is up to clients to determine the current offset between a source and target time zone when converting.
4. When saving a meeting, automatically calculate the time zone when the time zone has not been manually selected. Otherwise, allow manual selection override.
5. For existing meetings, a data migration is required. Use the google's time zone API to populate the field on upgrade. This is a major upgrade and we should make the UX for updating the meetings really solid. The migration can be punted until a serveradmin both enables the time zone API with the google key and selects a setting in the UI.
6. Expose the time zone field in the semantic API.

This approach should not break anything for existing clients, and should not require anyone to change their actual meeting times. 

All that said, virtual-only servers are the only servers at present that need this feature, for example virtual-NA.org. Virtual NA will probably have to fill out the time zone for every meeting manually, because there is no location information to go off of for the time zone. We could probably add the field without a data migration in a first version, but crouton would need to be changed to do time zone calculations when the feature is enabled.

jbraswell:
First iteration should look like this:

Root server
1. Add column to database.
2. Add drop down to meeting editor.
3. Add field to GetSearchResults
4. Add feature toggle to GetServerInfo

Crouton
1. Check feature toggle in GetServerInfo
2. When feature toggle is on, so time zone conversions

Future iterations can handle the migration path, and this will not impact anyone else - there is no breaking change right now.

alanborning:
How do the valid values for the time zone column interact with daylight savings time? For example both EST and EDT are in the time-zones database, so 2) above would say they are both valid values for the column. Unfortunately I'm not seeing a great solution for this. If someone populates the field for a Boston meeting in the summer, they ought to put EDT. Is EDT still used in the winter, and people just have to fix it up mentally? Or use EST even in the summer? That doesn't seem right either.

[...]

PR #344, opened 2020-09-21
Add time zone support behind feature flag

pjaudiomv:
ref https://github.com/bmlt-enabled/bmlt-root-server/issues/343


main_server/local_server/server_admin/c_comdef_admin_main_console.class.php
```diff
@@ -1682,6 +1682,22 @@ public function return_single_meeting_basic_template()
...
                     $ret .= '<div class="clear_both"></div>';
                 $ret .= '</div>';
+
+            if (isset($this->my_localized_strings['meeting_time_zones_enabled'])) {
+                $ret .= '<div class="bmlt_admin_one_line_in_a_form clear_both">';
+                $ret .= '<span class="bmlt_admin_med_label_right">'.htmlspecialchars($this->my_localized_strings['comdef_server_admin_strings']['meeting_editor_screen_meeting_time_zone_label']).'</span>';
+                $ret .= '<span class="bmlt_admin_value_left">';
+                $ret .= '<select id="bmlt_admin_single_meeting_editor_template_meeting_time_zone_select_input" class="bmlt_admin_single_meeting_editor_template_meeting_time_zone_select_input">';
+                $ret .= '<option value=""></option>';
+                foreach (DateTimeZone::listIdentifiers(DateTimeZone::ALL) as $tz) {
```
jbraswell, line 1692: I think we should store a static list of these in the code, because the return value of this function could differ between PHP versions. I there is an official tzinfo database, we shove those into an array in the root server code.

jbraswell: You can probably store the list in `my_localized_strings` or something, or as a default setting

pjaudiomv:
Todo: JS for meeting save, default timezone value. possibly remove flag for admin ui editor if js is too much of a pain

main_server/local_server/server_admin/c_comdef_admin_main_console.class.php
pjaudiomv: this totes made sense, added to localized strings.

jbraswell:
I think this is done, but I want to do the following before I merge.

- [ ] Test a fresh installation.
- [ ] Add support for this to crouton, and test.
- [ ] Limit only to the time zones needed by `virtual-na.org` at first... because they'll be manually editing their data, and it's going to be really tough when there are that many options in the list.

I don't think I want to recommend that anyone other than `virtual-na.org` use this feature until the following is true
* Time zone is automatically populated when a meeting is saved by querying the google maps time zone API, disabling manual time zone selection.
* There is a way to automatically populate the time zone field for all existing meetings.

jbraswell:
At this point I think we just need to test that a fresh install properly adds the column.

pjaudiomv:
I just tested on new install and it added the column fine https://nnerna.charlestonna.org/main_server/

PR #347, opened 2020-10-18
WIP: slowly migrate time zones on each request

jbraswell:


jbraswell:
TODO

- [ ] validate geocoding by city/state will work for international servers
- [ ] validate this fails fast and doesn't hang if we take tz-api down

Issue #358, opened 2020-12-10
When Virtual Meeting Link is provided, prompt user to fill out Virtual Meeting Additional Info


jbraswell:
Any time the user attempts to save with a non-empty Virtual Meeting Link field, and the Virtual Meeting Additional Info field is empty, we should ask them if they're sure, and explain why it's a good idea to at a minimum put the Zoom ID, password, or similar information in that field. It's fine to call out zoom specifically, let's be practical.

radius314:
I started working on this.  Strange that the description of what the field is is actually the value

![image](https://user-images.githubusercontent.com/1850811/101969069-65244e00-3bf0-11eb-8fc1-b465174a0a57.png)

jbraswell:
I think Virtual Meeting Link OR Phone Dial-In Information. 

There are a ton of "NA by phone" meetings out there. Idk if any of them are using BMLT presently... but would like to make them "first class citizens" along with video-based virtual meetings just in case, if it's not too much work.

alanborning:
Easy enough to do it either way.  So I'll make it an "OR".

alanborning:
Closed with 45a63b4545e5f169a3fcf3a84507754108db5e67.

PR #365, opened 2020-12-12
WIP: Adding vm link info validation

radius314:
https://github.com/bmlt-enabled/bmlt-root-server/pull/365/files?diff=unified&w=1

### DIFF
--- a/main_server/local_server/server_admin/server_admin_javascript.js
+++ b/main_server/local_server/server_admin/server_admin_javascript.js
@@ -293,19 +293,19 @@ function BMLT_Server_Admin()
     ****************************************************************************************/
     this.toggleAccountInfo = function () {
         this.m_account_panel_shown = !this.m_account_panel_shown;
-        
+
         var the_disclosure_div = document.getElementById('bmlt_admin_user_account_disclosure_div');
         var the_account_info_div = document.getElementById('bmlt_admin_user_account_wrapper_div');
-        
+
         if ( this.m_account_panel_shown) {
             var email_field = document.getElementById('bmlt_admin_user_email_input');
             var description_field = document.getElementById('bmlt_admin_user_description_textarea');
             var password_field = document.getElementById('bmlt_admin_user_account_password_input');
-            
+
             this.handleTextInputBlur(email_field);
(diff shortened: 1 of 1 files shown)


main_server/local_server/server_admin/server_admin_javascript.js
```diff
@@ -1392,7 +1392,22 @@ function BMLT_Server_Admin()
...
+
+    this.alertIssue = function ( message ) {
+        alert(message);
+    };
+
+    this.validateSave = function ( in_meeting_id ) {
+        var virtual_meeting_link = document.getElementById('bmlt_admin_single_meeting_editor_' + in_meeting_id + '_meeting_virtual_meeting_link_text_input');
+        var virtual_meeting_additional_info = document.getElementById('bmlt_admin_single_meeting_editor_' + in_meeting_id + '_meeting_virtual_meeting_additional_info_text_input');
+        if (virtual_meeting_link.value !== null && virtual_meeting_additional_info.value == null) {
+            alert("Virtual Meeting Additional Information field must be set.");
```
jbraswell, line 1404: I'm not sure the additional info field should be required, but I do think it should be _strongly suggested_. Thoughts?

jbraswell (changes requested):

main_server/local_server/server_admin/server_admin_javascript.js
pjaudiomv: I don't think it should be required either but I barely care so could go either way :)

radius314: Ok, this was an attempt at an implementation of this ticket.  How did you envision it would be?  https://github.com/bmlt-enabled/bmlt-root-server/issues/358.

jbraswell: The ticket says "we should ask them if they're sure, and explain why it's a good idea to at a minimum put the Zoom ID, password, or similar information in that field. It's fine to call out zoom specifically, let's be practical."

"We should ask them if they are sure"

If you guys both feel strongly that it should be required, then I'm fine with that - the UI should indicate clearly _why_ it is required, though.

PR #367, opened 2020-12-13
adding jquery

radius314:


radius314:
This results in an extra 25k downloaded loaded over the wire.

radius314:
I believe we mostly need this right now for selector capability so I'm using the `slim` version.  If we want to use other features we could always swap it out later.

https://stackoverflow.com/questions/35424053/what-are-the-differences-between-normal-and-slim-package-of-jquery

jbraswell (approved):

PR #371, opened 2020-12-26
fix issue with which meeting formats are selected for a NAWS export

### LINKED ISSUE #361: In NAWS Export, list virtual formats first
The goal here is to make sure that the virtual formats (VM, HY) make it into one of the five formats columns in the NAWS Export.

alanborning:
Should close #361

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,7 @@ For instructions on installing the root server, see [Installing a New Root Serve
 - Added initial support for time zones, hidden behind feature flag.
 - Add a "publish" checkbox to the NAWS import functionality to allow new meetings to be either published or unpublished.
 - Fixed an issue where map searches using `geo_width` could return incomplete results when filtered by service body.
-- Fixed an issue with which meeting formats are selected for a NAWS export. NAWS exports are limited to 5 formats.
-  `auto-config.inc.php` can now specify an optional variable `$naws_export_formats_at_front` that is an array of formats
-  that should be exported before others. If not present, the default value is `['VM', 'TC', 'HYBD', 'W', 'M', 'GL']`.
+- Fixed an issue with which meeting formats are selected for a NAWS export. NAWS exports are limited to 5 formats. `auto-config.inc.php` can now specify an optional variable `$naws_export_formats_at_front` that is an array of formats that should be exported before others. If not present, the default value is `['VM', 'TC', 'HYBD', 'W', 'M', 'GL']`.
 
 ***Version 2.15.6* ** *- September 7, 2020*
 - Added support for the `OLM` NAWS Format code for online meetings.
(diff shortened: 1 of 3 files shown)


main_server/server/c_comdef_server.class.php
```diff
@@ -3038,7 +3038,12 @@ public static function GetLocalStrings( $in_lang_enum = null  ///< An enumeratio
                 if (isset($default_timezone) && $default_timezone) {
                     date_default_timezone_set($default_timezone);
                 }
-            
+                // Due to NAWS database limitations, only 5 formats at most can be exported. auto-config.inc.php can specify an optional variable $naws_export_formats_at_front
+                // that is an array of formats that should be exported first, to make sure they are included. This is particularly important for VM (Virtual Meeting),
+                // TC (Temporarily Closed) and HYBR (Hybrid), since these are essential and we don't want them falling off the end. If this variable isn't set in
+                // auto-config.inc.php, give it a default value that includes those plus a few others. If the meeting has any of these formats, they are exported
+                // first, in the order they appear in naws_export_formats_at_front.
+                c_comdef_server::$server_local_strings['naws_export_formats_at_front'] = isset($naws_export_formats_at_front) ? $naws_export_formats_at_front : ['VM', 'TC', 'HYBR', 'W', 'M', 'GL'];
```
radius314, line 3046: Is this something that we would ever want someone to override?

Also, what happens if the format code is different for another language here?

jbraswell: I believe these are the NAWS mappings, not the translated BMLT formats, so translation is not an issue.

I don't think overridability is important here.

jbraswell: But... if the PR has already built in the ability to override, I think it's fine to leave it.

alanborning: Right, these are NAWS mappings.  Regarding the ability to override, it does seem very unlikely anyone would want to change it. The first 3 (VM, TC, and HYBD) seem obvious and noncontroversial. Since it's programmed to allow additional preferences, I included Women, Men, and LGBT (with lower priority for inclusion, but since you get 5 these will always be included as well).  For these I can imagine someone getting bent out of shape -- we had some long running discussion in my area about the BMLT format codes for LGBT meetings for example -- so I made it overridable so that someone could change it if they care about this.  For example it seems more useful to know if it's a Women's meeting than if it's a candlelight meeting.  Could leave off LGBT ... dunno.

README.md
```diff
@@ -50,6 +50,9 @@ For instructions on installing the root server, see [Installing a New Root Serve
 - Added initial support for time zones, hidden behind feature flag.
 - Add a "publish" checkbox to the NAWS import functionality to allow new meetings to be either published or unpublished.
 - Fixed an issue where map searches using `geo_width` could return incomplete results when filtered by service body.
+- Fixed an issue with which meeting formats are selected for a NAWS export. NAWS exports are limited to 5 formats.
```
jbraswell, line 53: I think you may have some unneeded newlines in this new comment. This should all really just go on one line.

jbraswell (approved):
My only comment is a nitpick about a newlines in the release notes. Feel free to merge after fixing the newline thing. Thank you!

Issue #357, opened 2020-12-10
Require "Virtual Meeting Link" or "Phone Meeting Dial-in Number" fields for virtual meetings


jbraswell:
Prerequisite: https://github.com/bmlt-enabled/bmlt-root-server/issues/355

If a user has designated the meeting as either Virtual (VM) or Hybrid (HY), do not allow them to save the meeting without entering a Virtual Meeting Link or Phone Number. Add good error messages to the UI explaining why.

jbraswell:
For first release it might be a good idea to _warn_ the user that they should fill out either the "Virtual Meeting Link" or "Phone Meeting Dial-in Number" rather than requiring.

alanborning:
Closed with 45a63b4545e5f169a3fcf3a84507754108db5e67.

PR #378, opened 2021-01-21
Additional data checks

### LINKED ISSUE #356: Do not allow users to save meetings without location information
Meetings without location information have always been a problem, but they've become a _huge_ problem since virtual meetings have become a thing.

Let's make sure at least a city/state OR zip code is filled in.

We'll probably need to poke at the international servers to make sure that city/state OR zip code will work for them. If it doesn't, we'll need to change the logic, but regardless - some location information _must_ be required.

alanborning:
Close #356, #357, and #358


radius314 (approved):

radius314:
already in master

PR #380, opened 2021-02-03
validation messages

radius314:


radius314:
image

radius314:
If you want to test unchecking a radio button you can run something like this in the JS console

```
$("#bmlt_admin_single_meeting_editor_2633_meeting_venue_type_virtual").prop('checked', false);
```

radius314:
I tried implementing jqueryvalidation, and it seemed overly complex for what we are doing.  The documentation for the site was intermittently available.

radius314:
I think we probably need a generic message near the button to let people know to scroll up.  Adding that.

radius314:
Open to suggestions on the text.

image

alanborning:
These messages overlap with the messages in #378, which has been there 2 weeks. Oops, didn't explicitly ask someone to review it .... just did.

alanborning:
Since #378 has considerably more changes, how about if we merge that one first, if it's ok. In penance for forgetting to explicitly request a review on it I will volunteer to fix up anything in this PR that would need fixing as a result.

radius314:
Sure I don't mind rebasing.

radius314:
I have re-integrated changes from #378.  There are probably a bunch of things that need to be done to clean this up and make the validation contexts a little clearer.

radius314:
image

radius314:
image

radius314:
image

alanborning:
I ran through all of the use cases and error conditions I could think of -- the functionality is all there and works correctly. Maybe there is a little more UI polishing that could be done, although I'm not sure what. But basically it seems ready to go.

Here are the cases I tested:
- from an older release, entered meetings with all 8 possible combinations of TC, VM, and HY; and then in this branch made sure the editor handled each of them correctly
- checked entering virtual or hybrid meetings with all 6 combinations of phone number, virtual meeting link, and virtual meeting additional info present or absent; made sure the errors and warnings were correct for all cases
- checked a bad URL in the virtual meeting link
- checked a missing address

radius314:
Or maybe we deploy to unstable and let a few people take a spin.

alanborning (approved):

PR #385, opened 2021-02-13
hitting rate limiting on pulling images for ecs containers

pjaudiomv:
yea so we are hitting rate limiting on pulling images for ecs as well, i added creds to travis


radius314:
Use nexus.

pjaudiomv:
Well need to create a vpc endpoint first

jbraswell (approved):

PR #387, opened 2021-02-13
Execution role

pjaudiomv:


pjaudiomv:
I thought the role was inherited from the service, but not when using private registries it needs to be explicit

jbraswell (approved):
Run terraform fmt?

PR #392, opened 2021-02-14
Migrating to public ecr from dockerhub

pjaudiomv:
This will solve rate limiting issues


jbraswell (approved):
Fix the title and squash merge?

Issue #362, opened 2020-12-10
Don't allow users to edit the `virtual_location` format types


jbraswell:
Prerequisite: https://github.com/bmlt-enabled/bmlt-root-server/issues/355

Requirements
1. Users should not be able to add additional formats to the `virtual_location` format_type_enum.
2. Users should not be able to delete formats assigned to the `virtual_location` format_type_enum.
3. Users should not be able to change the NAWS Format or format_type_enum of formats assigned to the `virtual_location` format_type_enum.

alanborning:
Closed with #370

alanborning:
Sorry, this was closed prematurely. The VM, TC, and HY formats are still visible in the Format Administration tab, and we need to make sure the user can't delete them, or add new formats marked with the corresponding NAWS formats.

radius314:
are you taking this one?

alanborning:
Yes, can do.

alanborning:
Closed with #394

PR #394, opened 2021-02-21
disable deleting or changing the NAWS format for HY, VM, and TC formats

alanborning:


alanborning:
Closes #362

alanborning:
Changed to WIP pending fixing the visual indication that the English key for TC, VM, and HY formats is read-only.

alanborning:
Fixing the visual indication that the English key for TC, VM, and HY formats is read-only is a very minor issue .... returning this to be an open PR. After it's merged I'll open a new issue for the visual indication.

jbraswell (approved):

PR #398, opened 2021-03-07
update debug dockerfile

pjaudiomv:


docker/Dockerfile-bmltdebug
```diff
@@ -1,10 +1,14 @@
...
 
-ENV PHP_INI_PATH "/usr/local/etc/php/php.ini"
+ENV PHP_INI_PATH /etc/php/7.2/apache2/php.ini
+ENV PHP_XDEBUG_ENABLED: 1
 
-RUN pecl install xdebug-2.7.2 \
-    && /usr/local/bin/docker-php-ext-enable xdebug \
-    && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" >> ${PHP_INI_PATH} \
+RUN apt-get update \
+    && apt-get -yqq install \
```
jbraswell, line 7: I like to put my &&s at the end of the previous lines instead 🤷‍♂️

jbraswell (approved):

docker/Dockerfile-bmltdebug
pjaudiomv: yea it does look cleaner that way

PR #410, opened 2021-04-10
add checks for venue type formats - must have a key in every language

alanb2718:


jbraswell (approved):

radius314:
Can you explain this feature?

alanb2718:
As an example, consider the HY format for Italian.  In English, the key HY is read-only; but you can change it in Italian.  (For NA Italia it's `Ib` for "Riunione ibrida".)  However, before this commit you could set all of the fields for Italian to null, save changes, and the Italian version of HY would simply be deleted.  That messes up badly for storing the venue type when using Italian.  So this change requires that you have **some** key -- so the format will be there in each language -- but you can change the key to something other than HY for languages other than English.

The migration code for 2.16.0 does make sure there is a format for HY, TC, and VM in all languages -- you'd have to perversely delete it after migration has run for this to arise.  I thought about a less restrictive version that would require the key just in the principal language -- but the principal language can be changed, so it seemed simpler to just require the key for all of them.

radius314:
Got it.

PR #411, opened 2021-04-15
Support teamspeak URLs for Virtual Meeting Link

jbraswell:


alanb2718 (approved):
This seems fine. If we want to extend the list of protocols much more, we could have it read an optional variable in `auto-config.inc.php`, rather than hardwiring everything.

jbraswell:
Making it a configuration option in the future is a good idea!

PR #412, opened 2021-04-17
wip: for adding venue_type to get search results

pjaudiomv:


pjaudiomv:
this would populate a  new db entry for venue_type as well, at some point i guess there should be a data migration based off existing logic but maybe not yolo

pjaudiomv:
this is just brute forcing things now, the validation logic hurts my brain.

pjaudiomv:
not the venue validation logic, the js meeting validation stuff

jbraswell:
I don't think I'm comfortable making an API change like this yet.

In the future (like, months at least) we will need to do something like this, but I want to slow roll it... because we will not be able to undo a change like this, and I want it to be well thought out and agreed with complete consensus among all of the root server contributors. In the meantime, let's let the world settle into the changes introduced in 2.16.0, and code the plugins around the format designators for venue type.

pjaudiomv:
👌

PR #420, opened 2021-05-09
Update debugging instructions, change xdebug port to 10000

jbraswell:


alanb2718 (commented):
Looks like this is already merged.  But I ran through the instructions again, and everything checks out.

PR #422, opened 2021-06-03
fix GetSearchResults; checks for venue type formats; other bug fixes

### LINKED ISSUE #348: Setup step 2 broken link
When setting up a new server the step two link [you need to register with Google](https://bmlt.app/google-maps-api-keys-and-geolocation-issues/) is a 404

### LINKED ISSUE #416: switcher=GetSearchResults&get_used_formats ignores the lang_enum parameter
`get_used_formats` seems to ignore the `lang_enum` parameter, and instead returns a list of formats just based on the `lang_enum` for the meeting.  For example, if the meeting has `lang_enum` set to `'en'`, then the call will return the keys for the English formats, even if you call it with `switcher=GetSearchResults&get_used_formats&lang_enum=pt`.

This causes problems with crouton, which tries to display the format names and descriptions using the keys from the `lang_enum` for the meeting in the database, and the names and descriptions using the language parameter to the crouton shortcode. For example this URL displays meetings in English: https://bmlt.wszf.org/meetings/?service_body=3.  So far so good.

But see https://bmlt.wszf.org/meetings/?service_body=3&language=pt-BR.  The format names and descriptions are indeed in Portuguese.  But the keys are for the English versions of the formats -- if it happens to be the same as the Portuguese key, the format shows up, and otherwise you may not see it.  Or if the key exists in both English and Portuguese, but for different formats, you get the English key and the Portuguese description. See for example the Step Brothers meeting Wednesday at 7pm (at least in the current schedule), which has the English key M (men).  In Portuguese, the format for a women's meeting has the key M (Mulheres).

### LINKED ISSUE #419: Incorrect URL in BMLT when API key is incorrectly configured
The warning link displayed at the top of the BMLT admin page after logging in, when there is an incorrectly configured API key is a 404: [https://bmlt.app/google-maps-api-keys-and-geolocation-issues/](url)
Screen Shot 2021-05-09 at 7 19 28 AM

It should be: [https://bmlt.app/google-api-key/](url)

### LINKED ISSUE #421: Language Consistency - Update Meeting World IDs
The button for the server admin to update world committee codes is labeled "Update Meeting World IDs" whereas the field on the Basic tab in the individual meeting information is "World Committee Code." 

I think the language should be consistent across the root server. So maybe the button to import the spreadsheet from NAWS should say "Update World Committee Codes" ??  And the tab name should probably also be "Update World Committee Codes from NAWS Spreadsheet" and the text next to the "choose file" box should be consistent as well.

alanb2718:
This is intended to be a new minor release 2.16.1.  Closes #348, closes #416, closes #419, closes #421.


pjaudiomv:
does tomato have same issue do we know

alanb2718:
Regarding #421 and tomato, tomato does handle the `lang_enum` parameter for `GetSearchResults` correctly already. I did notice another odd thing: the root server was also ignoring the `lang_enum` parameter another place when returning the result.  So for example if you had requested results in Portuguese, the returned data would now use the Portuguese keys for the formats in the return value -- that fixes the bug.  But it would still say `lang_enum='en'`.  This seemed weird so I fixed that too.  That all seems to work -- but I don't actually see anyplace that uses this `lang_enum` value, so it doesn't seem to matter.  I wasn't able to figure out how to check what tomato does for this part to do a comparison.

jbraswell (approved):

PR #424, opened 2021-06-29
add warnings for unnecessary information for virtual or in-person meetings

alanb2718:
- add warnings for in-person meetings with unnecessary information in virtual info fields
- add warnings for virtual meetings with unnecessary information in location name or address


main_server/local_server/db_connect.php
```diff
@@ -258,7 +258,7 @@ function DB_Connect_and_Upgrade()
                     $next_id = $next_id['next_id'];
                     $langs = array("en", "es", "fa", "fr", "it", "pl", "pt", "sv");
                     foreach ($langs as $lang) {
-                        $sql = "INSERT INTO `$table` (`shared_id_bigint`, `key_string`, `icon_blob`, `worldid_mixed`, `lang_enum`,`name_string`, `description_string`, `format_type_enum`) VALUES ($next_id, 'TC', NULL, NULL, '$lang', 'Temporarily Closed', 'Facility is Temporarily Closed', 'O')";
+                        $sql = "INSERT INTO `$table` (`shared_id_bigint`, `key_string`, `icon_blob`, `worldid_mixed`, `lang_enum`,`name_string`, `description_string`, `format_type_enum`) VALUES ($next_id, 'TC', NULL, NULL, '$lang', 'Temporarily Closed Facility', 'Facility is Temporarily Closed', 'O')";
```
jbraswell, line 261: Generally when doing db migrations, unless data corruption was introduced, it is best practice to leave old migrations alone and just manipulate data in the new migration. Any reason that wouldn't work here?

jbraswell (approved):
Approved but with a question about migrations. Please use your best judgment on how to proceed, and then merge. Thank you!

main_server/local_server/db_connect.php
alanb2718: That makes sense!  I undid this change (which will now be made by the next migration function instead).

PR #427, opened 2021-09-04
first pass moving tf to mvana

pjaudiomv:


pjaudiomv:
ok this is ready and applied, i destroyed the old terraform with exception of ECR

jbraswell (approved):

PR #432, opened 2021-10-15
fix NAWS import bugs

alanb2718:
allow marking deleted meetings when logged in as something other than serveradmin; fix NAWS import permissions bug


main_server/local_server/server_admin/c_comdef_admin_ajax_handler.class.php
```diff
@@ -334,18 +326,22 @@ private function MarkDeletedMeetingAsHandled($meeting_id)
                 foreach ($obj_array as $change) {
                     if ($change instanceof c_comdef_change and $change->GetChangeType() === 'comdef_change_type_delete' and !$change->GetAfterObject()) {
                         $meeting = $change->GetBeforeObject();
-                        if ($meeting->GetWorldID() === 'deleted') {
-                            return false;
+                        if (!$isServerAdmin && !in_array(intval($meeting->GetServiceBodyID()), $userServiceBodyIDs)) {
+                            // meeting was under some other service body
+                            return 'not_found';
```
jbraswell, line 331: I think the changes to this file look really good. The only suggestion I'd make is that now that we are returning these `'not_found'`, `'marked'`, and `'not_marked'` strings from a function, maybe I'd store their values in variables. Really, an enum object would be ideal, but the versions of PHP we are using don't have a proper enum.

You can abuse a class to do the same thing.

```
abstract class SomeEnum {
    const NOT_FOUND = 'not_found';
    const MARKED = 'marked';
    const NOT_MARKED = 'not_marked';
}

// and then you could replace all instances of the string with...
SomeEnum::NOT_FOUND
SomeEnum::MARKED
SomeEnum::NOT_MARKED
```

main_server/server/c_comdef_server.class.php
```diff
@@ -3077,18 +3077,15 @@ public static function GetUserServiceBodies( $in_user_id = null   ///< The ID of
...
                     $editors = $service_body->GetEditors();
-                    
                     if (is_array($editors) && count($editors)) {
                         if (in_array($in_user_id, $editors)) {
                             $is_editor = 'editor';
                         }
                     }
-                    
-                    if ($service_body->GetPrincipalUserID() == $in_user_id) {
+                    if ($service_body->IsUserInServiceBodyHierarchy(self::GetCurrentUserObj())) {
```
jbraswell, line 3085: I could be wrong, but I think `IsUserInServiceBodyHierarchy` will return true if the user is not an owner, but only an editor of a parent service body. I don't think that is what we want. Only "owner" permissions should recurse downwards through the tree, not "editor" permissions.

jbraswell (changes requested):

main_server/server/c_comdef_server.class.php
alanb2718: Right, `IsUserInServiceBodyHierarchy` will also return true if the user is not an owner, but only an editor of a parent service body.   But permission to edit meetings does recurse downward through the tree, so for consistency it seems reasonable to do it here also.  For example, if I create a new user `zonewide` and add it to the Full Meeting List Editors for the zone, then `zonewide` can edit a meeting whose service body is an area in the zone.  Seems plausible for `zonewide` to be able to import NAWS world IDs also ....

PR #440, opened 2021-11-20
add key_strings filter to GetFormats

jbraswell:


alanb2718 (commented):
I tested this for various combinations of formats and returned data types.  It works fine for returning csv or xml data, but not for Simple HTML (Table) or Simple HTML (Block Elements).  For the latter it didn't filter out any formats.  For example this gives all the formats rather than just VM:

`http://localhost:8000/main_server/client_interface/simple/?switcher=GetFormats&key_strings[]=VM`

Of course, who knows if anyone would ever use this returned data type and filter together, but it probably should work (or explicitly give an error).

jbraswell:
> I tested this for various combinations of formats and returned data types. It works fine for returning csv or xml data, but not for Simple HTML (Table) or Simple HTML (Block Elements). For the latter it didn't filter out any formats. For example this gives all the formats rather than just VM:
> 
> `http://localhost:8000/main_server/client_interface/simple/?switcher=GetFormats&key_strings[]=VM`
> 
> Of course, who knows if anyone would ever use this returned data type and filter together, but it probably should work (or explicitly give an error).

Thanks for finding that. Just implemented and tested for /simple/ endpoints.

alanb2718 (approved):
Works now!

Issue #460, opened 2022-05-04
add show_all to GetFormats in Semantic interface


pjaudiomv:
with the addition of `show_all` parameter added to GetFormats we should add a checkbox to toggle it on or off to the semantic interface

https://github.com/bmlt-enabled/bmlt-root-server/pull/459

pjaudiomv:
Resolved by https://github.com/bmlt-enabled/bmlt-root-server/pull/461

PR #461, opened 2022-05-04
add show_all formats to semantic ui

pjaudiomv:


pjaudiomv:
closes https://github.com/bmlt-enabled/bmlt-root-server/issues/460

jbraswell (approved):

Issue #481, opened 2022-08-30
Build/CI: Separate docker prod and debug images


jbraswell:
The `make dev` target should run a debug image, and only the debug image should have xdebug installed and configured.

pjaudiomv:
just to clarify the CI deployed image should not have xdebug, is that the ask

jbraswell:
Right - we don't want to run xdebug in a publicly accessible server.

PR #489, opened 2022-08-30
add debug dockerfile

pjaudiomv:
in support of https://github.com/bmlt-enabled/bmlt-root-server/issues/481


jbraswell:
Would be better if the debug Dockerfile inherited from the base Dockerfile.

Might make sense to do a multi-stage build and target a specific stage.

Also, the docker-compose file will need to change.

PR #513, opened 2022-09-01
rework make targets

jbraswell:


Makefile
```diff
@@ -15,51 +15,59 @@ else
...
-	curl -sLO https://github.com/bmlt-enabled/crouton/releases/latest/download/croutonjs.zip
-	mkdir -p src/legacy/client_interface/html/croutonjs
-	unzip croutonjs.zip -d src/legacy/client_interface/html/croutonjs
-	rm croutonjs.zip
-	rm src/legacy/client_interface/html/croutonjs/*.html
-	rm src/legacy/client_interface/html/croutonjs/*.json
-	rm -rf src/legacy/client_interface/html/croutonjs/examples
+.PHONY: ecr-login
+ecr-login: ## Authenticates to ECR
+	aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/t5y4k5q3
```
pjaudiomv, line 63: we can make this public.ecr.aws/bmlt
this was the registry name before the accepted our alias and we moved everything in small span

Makefile
```diff
@@ -15,51 +15,59 @@ else
...
+	rm -rf src/legacy/client_interface/html/croutonjs/examples
+
+$(ZIP_FILE): $(VENDOR_AUTOLOAD) $(CROUTON_JS)
+	mkdir -p build
+	cp -r src build/main_server
+	cd build && zip -r $(shell basename $(ZIP_FILE)) main_server
+	rm -rf build/main_server
 
-.PHONY: help
 help:  ## Print the help documentation
```
pjaudiomv, line 41: can we move this to be the top target. so if you run make with no target or phony it displays the help

PR #533, opened 2022-09-03
read laravel db settings from auto-config.inc.php

jbraswell:


src/config/database.php
```diff
@@ -43,15 +41,15 @@
         'mysql' => [
             'driver' => 'mysql',
             'url' => env('DATABASE_URL'),
-            'host' => env('DB_HOST', '0.0.0.0'),
+            'host' => $dbServer ?? env('DB_HOST', '0.0.0.0'),
```
pjaudiomv, line 44: nice we can use php functions 7 and greater

Issue #12, opened 2018-07-07
Inconsistent data response on semantic admin endpoint.


radius314:
If you make a request to the semantic admin as such:

`/local_server/server_admin/json.php?admin_action=get_permissions`

And the user you're logging in as has only one service body the response will not be an array.  However if there are more than one it will be.   It's my belief that it should always return a consistent model (in this case, always return an array).  I realize this might break other clients and things will have to be adjusted accordingly.  I will be working around this issue in the interim to not block development on this issue found: https://github.com/radius314/yap/issues/89

One service body (no array):
![https___bmltserver_org_main_server__local_server_server_admin_json_php_admin_action_get_permissions](https://user-images.githubusercontent.com/1850811/42414117-8b544950-81fc-11e8-820b-0796aa4a5d90.png)

More than one service body (array):
![https___bmltserver_org_main_server__local_server_server_admin_json_php_admin_action_get_permissions](https://user-images.githubusercontent.com/1850811/42414122-9904d7ea-81fc-11e8-9cd3-c18e2d5b7f94.png)

LittleGreenViper:
You are correct. There are probably a number of things like this throughout the repo.
The issue is that fixing them (like the gKey issue) may break stuff that is out there.
It will take time to analyze this, and time is something that's in short supply for me, these days.

radius314:
Worked around it, so it’s non-urgent.

jbraswell:
I agree with this issue, but I'm afraid fixing it will break existing contracts for consumers of this API. Can we enumerate the repos of code that use this API, and check the for compatibility? If we can validate it won't break anything downstream, I'll take a look at fixing it.

radius314:
I actually don't care about it anymore and also am worried about Yap breaking at this point.

PR #536, opened 2022-09-04
rewrite GetServiceBodies

jbraswell:


src/app/Http/Controllers/SwitcherController.php
```diff
@@ -0,0 +1,75 @@
...
+        $switcher = $request->input('switcher');
+
+        if (in_array($switcher, ['GetServiceBodies'])) {
+            $response = null;
+
+            if ($switcher == 'GetServiceBodies') {
+                $serviceBodies = $this->getServiceBodies($request);
+                if (!$serviceBodies->exists()) {
+                    // TODO look at clients and see if this is really necessary... I would much rather
+                    // TODO just return an empty list
```
pjaudiomv, line 33: I can check on this, I wouldn't be surprised if some of them don't even check for empty service bodies in a server

pjaudiomv: just checked, bread would not be affected from api if it were an empty list

pjaudiomv: it doesnt appear yap would be either

pjaudiomv: note that they would have separate issues with a server with no service bodies but nothing to do with empty response changing to empty list

jbraswell: I think I recall some client checking for {} from GetSearchResults. Just couldn't remember.

pjaudiomv: ah ok I specifically just checked GetServiceBodies, I think your right about GetSearchResults, ill see what client does that

pjaudiomv: yes both yap and crouton check for `{}` in GetSearchResults

jbraswell: Nice, we should open PRs to both of them to work with both {} and [].

pjaudiomv: and then what put a requirement on 3.0 requiring certain versions of yap/crouton

jbraswell: yes

pjaudiomv: Cool

pjaudiomv: Is this still relevant https://github.com/bmlt-enabled/bmlt-root-server/issues/12

jbraswell: Yes, that's what we're talking about. We should fix the clients to to work with an object or a list for now, because 3.0 is going to only return lists.

PR #605, opened 2022-09-19
mv license

pjaudiomv:


pjaudiomv:
@alanb2718  does this look legit(ish) or should we leave someones name. Like the bmlt isnt copyrighted but want to know the best way to set it to potentially protect us from being sued

alanb2718:
I would make it just `Copyright (c) 2018-present The BMLT Project` or something on that order.  As far as I know, it's OK to have an unincorporated association as the copyright holder, and it seems better than using people's names.

pjaudiomv:
> I would make it just `Copyright (c) 2018-present The BMLT Project` or something on that order. As far as I know, it's OK to have an unincorporated association as the copyright holder, and it seems better than using people's names.

Ok I just changed it to `The BMLT Enabled Community`

PR #613, opened 2022-09-23
implement delete service body rest endpoint

jbraswell:


src/app/Http/Controllers/Admin/ServiceBodyController.php
```diff
@@ -79,6 +80,14 @@ public function update(Request $request, ServiceBody $serviceBody)
 
     public function destroy(ServiceBody $serviceBody)
     {
-        //
+        if ($serviceBody->children()->exists() || $serviceBody->meetings()->exists()) {
```
pjaudiomv, line 83: nooice

PR #626, opened 2022-09-25
turn off fail fast to allow all matrix tests to run

pjaudiomv:


pjaudiomv:
```
If fail-fast is set to true, GitHub will cancel all in-progress and queued jobs in the matrix if any job in the matrix fails. This property defaults to true.
```

jbraswell (approved):

PR #699, opened 2022-10-06
self host swagger ui

jbraswell:


src/app/Http/Controllers/Admin/SwaggerController.php
```diff
@@ -0,0 +1,22 @@
...
+    public function openapi(Request $request)
+    {
+        $server = new \stdClass;
+        $server->url = rtrim($request->getBaseUrl(), '/') . '/';
+        $server->description = 'this server';
+
+        $json = json_decode(\File::get(storage_path('api-docs/api-docs.json')));
+        $json->servers = [$server];
+
+        return new JsonResponse($json, headers: ['content-type' => 'text/json']);
```
pjaudiomv, line 20: we should make this `application/json` no one uses `text/json` anymore

pjaudiomv: na

Issue #423, opened 2021-06-04
font display bug in Meeting History display


alanb2718:
There is a problem in the Meeting History with displaying letters with diacritical marks (accents etc).  See these screenshots of a test meeting history in English and French.  (Same thing happens in other languages with diacritical marks.)

![history-example-english](https://user-images.githubusercontent.com/918773/120843246-bd466800-c522-11eb-8602-4c8dd966358c.jpg)

![history-example-french](https://user-images.githubusercontent.com/918773/120843283-c6373980-c522-11eb-868a-2f91572b1e2b.jpg)

pjaudiomv:
This is no longer an issue in 4.0.0

<img width="624" height="185" alt="Image" src="https://github.com/user-attachments/assets/27d1deef-94a9-4089-9be9-8aef97caae1a" />

PR #717, opened 2022-10-11
fix for change history display bug #423

pjaudiomv:
this function is only used by change history on admin ui


pjaudiomv:
closing will fix in 3.0 UI

PR #762, opened 2022-10-20
return user_id with token

jbraswell:
The UI can use the user id to look up the current user at `GET /users/{id}`.


GarrettWeinberg:
is this just returning the userID?

jbraswell:
Yeah, it's just returning the `user_id`. I thought about returning the entire user object, but a user's attributes can change over the lifetime of a token, so it's better to just give the UI the `user_id` so that it can look them up itself.

GarrettWeinberg (approved):

pjaudiomv (approved):

PR #787, opened 2022-10-23
add test for naws export with no world id for sb

pjaudiomv:


alanb2718:
Do we actually want this functionality?  I originally allowed doing a NAWS export of a service body with no world ID (the old code allows this as well).  It would be an odd thing to export for use by NAWS, but maybe someone else wants to get a dump of an area they created experimentally or something.  Why not just allow it?

pjaudiomv:
> Do we actually want this functionality? I originally allowed doing a NAWS export of a service body with no world ID (the old code allows this as well). It would be an odd thing to export for use by NAWS, but maybe someone else wants to get a dump of an area they created experimentally or something. Why not just allow it?

The old code did not allow this, at least not 2.16.2. I tested this on both sezf and nerna servers. I did a export then removed world id from an area and did another export and those meetings were gone. Interestingly if you try to do a naws export of a service body with no world id it just exports the whole server

jbraswell:
For this rewrite, I'd like to keep parity with the old code except in cases with obvious bugs. That way after people upgrade it'll be easier to track down any actual defects.

alanb2718:
I think the old server removes the command to export a NAWS dump in the UI, but it allows you to do it anyway (for example if you put together the URL in the semantic workshop).

jbraswell:
It may be worth retesting, but I think Patrick is actually right about this one. We saw it when comparing zonal exports pre and post rewrite.

pjaudiomv:
> I think the old server removes the command to export a NAWS dump in the UI, but it allows you to do it anyway (for example if you put together the URL in the semantic workshop). So if we are strictly maintaining parity, it **should** be allowed in this part of the code ...

yea theres a bug there, if you do an export (using semantic query) for a service body with no world if it exports the whole server

alanb2718:
Oh, you're right!  It does export the whole server!!  OK, let's forbid it.

PR #796, opened 2022-11-03
configure global state management

GarrettWeinberg:
- use useContext in place of mobx
- set logged in user name after successful login


src/resources/js/App.tsx
```diff
@@ -1,18 +1,23 @@
...
   return (
-    <ThemeProvider theme={theme}>
-      <BrowserRouter basename='/main_server/'>
-        <CssBaseline />
-        <Router />
-      </BrowserRouter>
-    </ThemeProvider>
+    <AppContext.Provider value={{ userName, setUserName }}>
+      <ThemeProvider theme={theme}>
+        <BrowserRouter basename='/main_server/'>
```
pjaudiomv, line 15: is this just temporary, not all servers start with `/main_server/`

jbraswell: You can parse the actual basename from `new URL(baseUrl)`. The baseUrl is global, defined in the base laravel template.

GarrettWeinberg: we will have to figure out something for that.

GarrettWeinberg: but otherwise it will default to `/` so we will need to find what the correct permalink is.

PR #811, opened 2022-11-08
Begin working on User page

Jneville0815:


src/resources/js/pages/Users.tsx
```diff
@@ -1,5 +1,66 @@
...
+    });
+  };
+
+  const getUsers = async () => {
+    try {
+      const allUsers = await RootServerApi.getUsers();
+      setUsers(allUsers);
+      console.log(allUsers);
+    } catch (error: any) {
+      console.log(error);
```
jbraswell, line 33: Look for examples of calling `RootServerApi.handleErrors`.

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
jbraswell, line 1: Probably delete this file.

src/resources/js/pages/Users.tsx
```diff
@@ -1,5 +1,66 @@
...
+
+  useEffect(() => {
+    getUsers();
+  }, []);
+
+  return (
+    <div>
+      <Box>
+        <FormControl>
+          <InputLabel id='select-label'>User</InputLabel>
```
jbraswell, line 45: Everwhere you add a string, add an english translation to `localization.ts`. Then you can import `strings` and use the translation.

src/resources/js/pages/Users.tsx
GarrettWeinberg: can do 
```
import {
Box,
FormControl,
...} from @mui/material
```

GarrettWeinberg: Just my preference tho

src/resources/js/pages/Users.tsx
```diff
@@ -1,5 +1,66 @@
...
 export const Users = () => {
-  return <div>{strings.usersTitle}</div>;
+  const [currentSelection, setCurrentSelection] = useState('create-new-user');
+  const [selectedUser, setSelectedUser] = useState<any>({});
+  const [users, setUsers] = useState<User[]>([]);
+
+  const handleChange = (event: SelectChangeEvent) => {
+    setCurrentSelection(event.target.value as string);
+    users.forEach((user) => {
+      if (event.target.value === user.username) {
```
jbraswell, line 19: Would it be better to use `user.id` for the event target?

src/resources/js/pages/Users.tsx
```diff
@@ -1,5 +1,84 @@
...
+import Select, { SelectChangeEvent } from '@mui/material/Select';
+import { User } from 'bmlt-root-server-client';
+import { useEffect, useState } from 'react';
+
+import RootServerApi from '../RootServerApi';
 import { strings } from '../localization';
 
 export const Users = () => {
-  return <div>{strings.usersTitle}</div>;
+  const [currentSelection, setCurrentSelection] = useState('-1');
```
jbraswell, line 10: Should you be passing the number `-1` instead of the string `'-1'`?

src/package-lock.json
```diff
@@ -1,8790 +0,0 @@
-{
-    "name": "src",
```
jbraswell, line 2: This file needs to be restored.

src/resources/js/localization.ts
```diff
@@ -4,16 +4,22 @@ export const strings = new LocalizedStrings({
...
     loginTitle: 'Login',
+    emailTitle: 'Email',
     usernameTitle: 'Username',
     passwordTitle: 'Password',
     rootServerTitle: 'Root Server',
     dashboardTitle: 'Dashboard',
     meetingsTitle: 'Meetings',
     serviceBodiesTitle: 'Service Bodies',
     usersTitle: 'Users',
+    userTitle: 'User',
```
jbraswell, line 15: Try to find translations for each of these strings from the existing translation files in the root server - I sent you the github url to those files in Slack.

src/resources/js/pages/Users.tsx
Jneville0815: The MenuItem value wants a string, so I just convert the ID to a string

src/resources/js/pages/Users.tsx
```diff
@@ -1,5 +1,84 @@
...
+  if (errorMessage) {
+    return <div>{errorMessage}</div>;
+  }
+
+  return (
+    <div>
+      <Box>
+        <FormControl>
+          <InputLabel id='select-label'>{strings.userTitle}</InputLabel>
+          <Select labelId='select-label' id='select' value={currentSelection} label='User' onChange={handleChange}>
```
jbraswell, line 52: the label here should use a translated string

jbraswell (approved):

PR #817, opened 2022-11-18
run js tests in ci

jbraswell:


.github/workflows/main.yml
```diff
@@ -82,6 +82,12 @@ jobs:
         run: |
           make lint-js
 
+      - name: make test-js
+        env:
+          LARAVEL_BYPASS_ENV_CHECK: '1'
```
jbraswell, line 87: `laravel-vite-plugin` tries to be helpful by blowing up if it detects the vite server is running in CI. The `LARAVEL_BYPASS_ENV_CHECK` env var disables this check.

src/package.json
```diff
@@ -5,7 +5,8 @@
         "build": "vite build",
         "lint": "tsc && prettier --check vite.config.ts resources/js && eslint vite.config.ts resources/js",
         "format": "prettier --write vite.config.ts resources/js",
-        "test": "vitest",
+        "test": "vitest run",
```
jbraswell, line 8: `vitest run` just runs the tests and exists.

`vitest` runs the tests, watches for changes and then continually re-runs the tests as files change.

GarrettWeinberg (approved):

Issue #307, opened 2020-06-25
Provide a way to search for multiple formats view a querystring without double square brackets []


dgershman:
Currently in order to search for multiple formats, you need to include the `[]` in the querystring.  For example `"&formats[]=57&formats[]=69"`.  This causes issues in Wordpress which uses square brackets to denote the beginning and end of a shortcode.  This was handled in the BMLT satellite plugin by using double square brackets `[[bmlt(xxxxx)]]`.  Most of the wordpress plugins out there that have been recently built do not support such a convention.  

Obviously apps would have to support both the current way and the proposed new way.  My suggestion would be to supply a charactered separated list like `"&formats=57~69"`.     

I chose `~` because it's one of the characters that does not need to be url encoded.  There are a small number of them.

pjaudiomv:
this has been completed for formats and services filters using a comma `,`

IE . `?switcher=GetSearchResults&formats=57,69`

IE. `?switcher=GetSearchResults&services=1009,1049`

PR #876, opened 2023-02-20
Allow for specifying a comma seperated string instead of array for filters

pjaudiomv:


pjaudiomv:
resolves https://github.com/bmlt-enabled/bmlt-root-server/issues/307

PR #881, opened 2023-03-20
Migrate $include_service_body_email_in_semantic setting

pjaudiomv:


docker/auto-config.inc.php
```diff
@@ -38,3 +38,4 @@
 $admin_session_name = 'BMLT_Admin'; // This is merely the 'tag' used to identify the BMLT admin session.
 $g_enable_semantic_admin = TRUE;
 $default_minute_interval = 5; // This sets the minutes interval for Start Time and Duration.
+#$include_service_body_email_in_semantic = TRUE;
```
rebellionit, line 41: we can remove this line

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 3.0.4 (TBD)
-* Allow for specifying a comma seperated string instead of array for the `services` and `formats` filters.
+* The `services` and `formats` filter parameters for the `GetSearchResults` API now accept a comma separated string (e.g. `services=1,-12,42`) in addition to the standard array syntax.
```
rebellionit, line 2: this actually works for GetServiceBodies with services filter as well as GetFormats with formats filter which is why i didnt explicitly only call out GetSearchResults. but this is fine, this is what most users want to know

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 3.0.4 (TBD)
```
pjaudiomv, line 1: lets remember to change this to todays date if your gonna release

PR #901, opened 2023-05-22
Format type UI

otrok7:
Group format types in the Admin UI.  The format types are expandable and configurable.  Some format types can be represented as radio buttons, and the system validates that one is selected-


src/app/Repositories/FormatRepository.php
```diff
@@ -83,7 +83,13 @@ public function getTemporarilyClosedFormat(): Format
             ->where('lang_enum', 'en')
             ->firstOrFail();
     }
-
+    public function getFormatsByFormatType($key, $lang = 'en'): Collection
+    {
+        return Format::query()
+            ->where('format_type_enum', $key)
+            ->where('lang_enum', 'en')
```
jbraswell, line 90: Did you mean to pass $lang here?

src/tests/Unit/FormatTypesTest.php
```diff
@@ -0,0 +1,18 @@
...
+class FormatTypesTest extends TestCase
+{
+    /**
+     * A basic unit test example.
+     *
+     * @return void
+     */
+    public function testDummy()
+    {
+        $this->assertTrue(true);
```
jbraswell, line 16: Should this file be removed? Or are you still working on it?

src/tests/Feature/GetServiceBodiesTest.php
```diff
@@ -59,7 +59,7 @@ private function allServiceBodiesInArray($expectedItems, $array): bool
                 'name' => $item->name_string,
                 'description' => $item->description_string,
                 'type' => $item->sb_type,
-                'url' => $item->uri_string,
+                'url' => $item->uri_string ?? '',
```
jbraswell, line 62: Why did this change?

jbraswell (changes requested):
Aside from the comments I left, my only real request is to choose a name for the new column other than "api key". I understand why you chose that name, but I find it confusing, because an "api key" is usually a type of authentication credentials for an API.

After those small changes, I'm okay with merging this.

docker/sample-bmlt-schema.sql
```diff
@@ -14,7 +14,32 @@
 /*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
 /*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
 /*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+DROP TABLE IF EXISTS `na_comdef_format_types`;
```
jbraswell, line 17: I don't think this schema file should be modified. The new database migrations should take over and create the new table.

otrok7:
Thanks for the review, and for catching the bug.  I changed api_key to api_enum, hope that is OK.  Can you have a look at why the tests are not running.  Locally, everything is fine, and the message is not helpful.  It says the file name doesn't match with the class name, but it does.  I need another set of eyes on this.

jbraswell:
> Thanks for the review, and for catching the bug.  I changed api_key to api_enum, hope that is OK.  Can you have a look at why the tests are not running.  Locally, everything is fine, and the message is not helpful.  It says the file name doesn't match with the class name, but it does.  I need another set of eyes on this.

The name api_enum seems fine, and sure, I'll take a look at the tests. Might be a day or two - busy weekend.

pjaudiomv:
somethings causing phpunit to crash from the last commit maybe in this diff https://github.com/bmlt-enabled/bmlt-root-server/compare/45abec92064fdd25883c230f56f6089b71d3055f...e9dfe714acf42cb1e16786b916fc491d5507c3a8 error messaging is less than ideal but think the file name error is a red herring

jbraswell (approved):

pjaudiomv (commented):
can we update the changelog

jbraswell (approved):

pjaudiomv (approved):

PR #913, opened 2023-07-14
upgrade to laravel 10, deprecate php 8.0

pjaudiomv:
https://laravel.com/docs/10.x/upgrade#database-expressions


pjaudiomv:
this relies on php spreadsheets next release, which is compatible with 8.3. master php spreadsheet works and tested https://github.com/PHPOffice/PhpSpreadsheet

jbraswell (approved):
thank you

PR #940, opened 2024-05-24
Add server_ip to GetServerInfo

pjaudiomv:


CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 * The `weekdays` filter parameter for the `GetSearchResults` API now accept a comma separated string (e.g. `weekdays=1,-2`) in addition to the standard array syntax.
 * Updated the NAWS import functionality for use by NAWS when importing virtual meetings, by importing the virtual meeting fields and time zone, and allowing the address and city to be blank.
 * Fixed a bug with trying to log in as a deactivated user.
-* Fixed an issue when using NAWS Export with certain very old data sets. 
+* Fixed an issue when using NAWS Export with certain very old data sets.
+* Added `server_ip` to `GeServerInfo`.
```
dgershman, line 6: typo?

src/app/Http/Controllers/Query/SwitcherController.php
```diff
@@ -547,7 +556,8 @@ private function getServerInfo($request): BaseJsonResponse
             'auto_geocoding_enabled' => legacy_config('auto_geocoding_enabled'),
             'commit' => config('app.commit'),
             'default_closed_status' => legacy_config('default_closed_status'),
-            'aggregator_mode_enabled' => legacy_config('aggregator_mode_enabled')
+            'aggregator_mode_enabled' => legacy_config('aggregator_mode_enabled'),
+            'server_ip' => $serverIp
```
dgershman, line 560: There may be a desire to opt-out of this since there is some risk associated with this.

src/app/Repositories/ServerAddressLookupRepository.php
```diff
@@ -0,0 +1,42 @@
...
+namespace App\Repositories;
+
+use App\Interfaces\ServerAddressLookupRepositoryInterface;
+use Illuminate\Http\Client\HttpClientException;
+use Illuminate\Support\Facades\Http;
+
+class ServerAddressLookupRepository implements ServerAddressLookupRepositoryInterface
+{
+    protected array $urls = [
+        "aHR0cDovL2NoZWNraXAuYW1hem9uYXdzLmNvbQ==",
```
dgershman, line 12: Any reason why you are using `http` instead of `https`?  Seems unnecessary to base64 encode.

src/app/Http/Controllers/Query/SwitcherController.php
dgershman: Also if there is an error maybe `$serverIp` should be empty or null, rather than returning the error in the response.  The error should be logged though.

jbraswell: This is in support of a hosted geocoding and time zone API that Patrick is working on. The idea is to use this endpoint to gather IPs from the servers in the tally for allowlisting. We do actually want to return the error here for troubleshooting, as we won't actually have access to most of the root servers.

src/app/Repositories/ServerAddressLookupRepository.php
jbraswell: Main use case for obfuscating stuff like this is to not show up in GitHub code searches.

pjaudiomv: totes should be https

PR #953, opened 2024-07-09
user delete

pjaudiomv:
Need to figure out refreshing user list after delete and handle translations

### DIFF
--- a/src/resources/js/components/UserDeleteForm.svelte
+++ b/src/resources/js/components/UserDeleteForm.svelte
@@ -11,51 +11,44 @@
   import { spinner } from '../stores/spinner';
   import { translations } from '../stores/localization';
 
-  export let selectedUser: User | null;
+  export let deleteUser: User;
 
   const dispatch = createEventDispatcher();
-  let checkboxChecked: boolean = false;
 
   const { form, errors } = createForm({
-    initialValues: { userId: selectedUser?.id, confirmCheckbox: false },
+    initialValues: { userId: deleteUser?.id, confirmed: false },
     onSubmit: async () => {
       spinner.show();
-      if (selectedUser) {
-        await RootServerApi.deleteUser(selectedUser.id);
-        dispatch('userDeleted', { userId: selectedUser.id });
-      }
+      await RootServerApi.deleteUser(deleteUser.id);
     },
     onError: async (error) => {
       console.log(error);
+      await RootServerApi.handleErrors(error as Error);
       spinner.hide();
     },
     onSuccess: () => {
       spinner.hide();
+      dispatch('deleted', { userId: deleteUser.id });
     },
     extend: validator({
       schema: yup.object({
-        confirmCheckbox: yup.boolean().oneOf([true], $translations.confirmYouMust)
+        confirmed: yup.boolean().oneOf([true], $translations.confirmYouMust)
       })
     })
   });
-
-  function closeModal() {
-    dispatch('close');
-  }
 </script>
 
 <form use:form>
   <div class="text-center">
     <ExclamationCircleOutline class="mx-auto mb-4 h-12 w-12 text-gray-400 dark:text-gray-200" />
-    <h3 class="mb-5 text-lg font-normal text-gray-500 dark:text-gray-400">{$translations.confirmAreYouSure}? <br />{selectedUser?.displayName}</h3>
+    <h3 class="mb-5 text-lg font-normal text-gray-500 dark:text-gray-400">{$translations.confirmAreYouSure} <br />{deleteUser?.displayName}</h3>
     <div class="mb-5">
-      <Checkbox id="confirmCheckbox" name="confirmCheckbox" bind:checked={checkboxChecked}>{$translations.confirmActionCantBeUndone}</Checkbox>
+      <Checkbox id="confirmed" name="confirmed">{$translations.confirmActionCantBeUndone}</Checkbox>
     </div>
-    <Button type="submit" color="red" class="me-2" disabled={!checkboxChecked}>{$translations.confirmYes}</Button>
-    <Button type="button" on:click={closeModal} color="alternative">{$translations.confirmNo}</Button>
+    <Button type="submit" color="red" class="me-2">{$translations.confirmYes}</Button>
     <Helper class="mt-2" color="red">
-      {#if $errors.confirmCheckbox}
-        {$errors.confirmCheckbox}
+      {#if $errors.confirmed}
+        {$errors.confirmed}
       {/if}
     </Helper>
   </div>
--- a/src/resources/js/components/UserDeleteForm.svelte
+++ b/src/resources/js/components/UserDeleteForm.svelte
@@ -1,0 +1,62 @@
+<script lang="ts">
+  import { Button, Helper, Checkbox } from 'flowbite-svelte';
+  import { createForm } from 'felte';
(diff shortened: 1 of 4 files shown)


pjaudiomv:
> Need to figure out refreshing user list after delete and handle translations

I split out to separate form and modal and added translations and dispatch event to filter user list

src/resources/js/components/UserDeleteModal.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
+  import { Modal } from 'flowbite-svelte';
+
+  import type { User } from 'bmlt-root-server-client';
+  import UserDeleteForm from './UserDeleteForm.svelte';
+
+  export let showUserDeleteModal: boolean;
```
jbraswell, line 7: You could just name this `showModal` inside of this file.

src/resources/js/components/UserDeleteModal.svelte
```diff
@@ -0,0 +1,19 @@
...
+  export let selectedUser: User | null;
```
jbraswell: This type shouldn't be nullable. The dialog will never be used without a `selectedUser`.

`export let selectedUser: User`

src/resources/js/routes/Users.svelte
```diff
@@ -14,6 +15,7 @@
   let users: User[] = [];
   let filteredUsers: User[] = [];
   let showModal = false;
+  let showUserDeleteModal = false;
```
jbraswell, line 18: A shorter name like `showDeleteModal` might be better here.

src/resources/js/components/UserDeleteForm.svelte
```diff
@@ -0,0 +1,69 @@
...
+        await RootServerApi.deleteUser(selectedUser.id);
+        dispatch('userDeleted', { userId: selectedUser.id });
+      }
+    },
+    onError: async (error) => {
+      console.log(error);
+      await RootServerApi.handleErrors(error as Error, {
+        handleValidationError: (error) => {
+          errors.set({
+            userId: (error?.errors?.userId ?? []).join(' ')
```
jbraswell, line 33: This API will never throw validation errors, so you don't need to handle them. You can just do `await RootServerApi.handleErrors(error as Error)`.

src/resources/js/components/UserDeleteForm.svelte
```diff
@@ -0,0 +1,69 @@
...
+        }
+      });
+      spinner.hide();
+    },
+    onSuccess: () => {
+      spinner.hide();
+    },
+    extend: validator({
+      schema: yup.object({
+        confirmCheckbox: yup.boolean().oneOf([true], $translations.confirmYouMust)
```
jbraswell, line 44: I would just call this `confirmed`. The fact that it came from a checkbox isn't relevant.

src/resources/js/components/UserDeleteForm.svelte
```diff
@@ -0,0 +1,69 @@
...
+
+<form use:form>
+  <div class="text-center">
+    <ExclamationCircleOutline class="mx-auto mb-4 h-12 w-12 text-gray-400 dark:text-gray-200" />
+    <h3 class="mb-5 text-lg font-normal text-gray-500 dark:text-gray-400">{$translations.confirmAreYouSure}? <br />{selectedUser?.displayName}</h3>
+    <div class="mb-5">
+      <Checkbox id="confirmCheckbox" name="confirmCheckbox" bind:checked={checkboxChecked}>{$translations.confirmActionCantBeUndone}</Checkbox>
+    </div>
+    <Button type="submit" color="red" class="me-2" disabled={!checkboxChecked}>{$translations.confirmYes}</Button>
+    <Button type="button" on:click={closeModal} color="alternative">{$translations.confirmNo}</Button>
```
jbraswell, line 62: I would get rid of this button and just let them use the x or click off of the modal to cancel.

src/resources/js/components/UserDeleteForm.svelte
```diff
@@ -0,0 +1,69 @@
...
+  const dispatch = createEventDispatcher();
+  let checkboxChecked: boolean = false;
+
+  const { form, errors } = createForm({
+    initialValues: { userId: selectedUser?.id, confirmCheckbox: false },
+    onSubmit: async () => {
+      spinner.show();
+      if (selectedUser) {
+        await RootServerApi.deleteUser(selectedUser.id);
+        dispatch('userDeleted', { userId: selectedUser.id });
```
jbraswell, line 25: Just call the event `deleted` so you can do `on:deleted`.

src/resources/js/components/UserDeleteForm.svelte
```diff
@@ -0,0 +1,69 @@
...
+</script>
+
+<form use:form>
+  <div class="text-center">
+    <ExclamationCircleOutline class="mx-auto mb-4 h-12 w-12 text-gray-400 dark:text-gray-200" />
+    <h3 class="mb-5 text-lg font-normal text-gray-500 dark:text-gray-400">{$translations.confirmAreYouSure}? <br />{selectedUser?.displayName}</h3>
+    <div class="mb-5">
+      <Checkbox id="confirmCheckbox" name="confirmCheckbox" bind:checked={checkboxChecked}>{$translations.confirmActionCantBeUndone}</Checkbox>
+    </div>
+    <Button type="submit" color="red" class="me-2" disabled={!checkboxChecked}>{$translations.confirmYes}</Button>
```
jbraswell, line 61: I think you can choose one of two approaches.

1. Keep the submit button enabled, and rely on yup error validation to show an error if they click submit when the checkbox isn't checked.
2. Get rid of the yup validation and keep the button disabled until the checkbox is checked.

I don't think you need both, and I'd probably do #1.

src/resources/js/routes/Users.svelte
```diff
@@ -56,6 +57,12 @@
     closeModal();
   }
 
+  function onDeleted(event: CustomEvent) {
+    const { userId } = event.detail;
```
jbraswell, line 61: It would be more readable to do `const userId = event.detail.userId;`

jbraswell (changes requested):

src/resources/js/components/UserDeleteForm.svelte
jbraswell: Also, I'd move the `dispatch` call to `onSuccess`.

src/resources/js/components/UserDeleteForm.svelte
```diff
@@ -0,0 +1,69 @@
...
+  export let selectedUser: User;
+
+  const dispatch = createEventDispatcher();
+  let checkboxChecked: boolean = false;
+
+  const { form, errors } = createForm({
+    initialValues: { userId: selectedUser?.id, confirmCheckbox: false },
+    onSubmit: async () => {
+      spinner.show();
+      if (selectedUser) {
```
jbraswell, line 23: If `selectedUser` can't be null, then you don't need this if condition.

src/resources/js/routes/Users.svelte
pjaudiomv: I was thinking that but wanted to be explicit for if we have like delte modal for service body or something

jbraswell: The filename tells you what the context is. Since this is in `Users.svelte`, it's implied that you're deleting users.

jbraswell (approved):

PR #973, opened 2024-07-24
several service body enhancements and fixes

jbraswell:
1. Removed the no admin option, because the api doesn't allow it.
2. Hide observer users from the list when they aren't already set on the service body.
3. Mark deactivated users in the admin select menu with `[DEACTIVATED]` text.
4. Make badges rounded and move their color selection to a function.
5. Fix styling of `Select` and `MultiSelect` so that they have the same background color as other form fields.


pjaudiomv:
what does 2. mean Hide observer users from the list when they aren't already set on the service body.

jbraswell:
It means observer users aren't listed in the admin or meeting editor list unless they are already assigned as an admin or editor on the service body.

pjaudiomv (approved):

PR #982, opened 2024-07-28
add service body form tests

pjaudiomv:


alanb2718 (changes requested):
Thanks for putting in these tests!  There are some failed tests and also lint is complaining about something, so that needs to be fixed before merging with main.

Here are some other questions and comments.

There are various style differences with the Users and Login tab tests.  How consistent should we be?  (Ideally it seems like we would be consistent, but also it isn't essential.)

There is a file `sharedDataAndMocks` for simulated test data like users.  I think it would be better to use the shared data file, and add service bodies to that.  We'll need the service bodies for some of the users tests, in particular the one that checks for an error message when trying to delete a user that is still associated with a service body.

These tests render the ServiceBodyForm directly, whereas the Users tab tests simulate logging in and picking the Users link from the navigation bar.  It seems like we should have tests that simulate clicking on the Users and the Service Bodies links, but we don't need all the tests to do that.  So rendering the form directly seems better, and I can refactor the Users tests and put the link checking test elsewhere.

These tests use `fireEvent` to mock clicking and typing.  In the other tests I used `@testing-library/user-event` which provides a higher-level API -- I like using `click` and `type` from that package for example. 

The Users and Login tests put all the mocks separate from the tests; and here some of the tests build in the mock (for example `'test Fields are disabled for non-admin users'`).  I think the test is cleaner to separate out the mocked method; but I'm sure there is room for debate on this!

These tests use for example `translations.getString('addServiceBody')` whereas the others just inline the string.  I prefer inlining the string, since it tests more of the functionality.  That does have the disadvantage that if you change a string used in a test, you have to change it both in localizations and the test; but there aren't that many.  Certainly for the language tests, I think we should write 

`await screen.findByText('Ungültiger Benutzername oder Passwort.');`

rather than 
`await screen.findByText(translations.getString('invalidUsernameOrPassword');`

since otherwise we're not checking that we switched to German.

So ... if we were trying to make things consistent between the tests, I'd change the tests I wrote to usually render components directly; and I'd change various other things in the ServiceBody test to be like the Users version.  I'd be willing to make a consistency pass through all of them if you would like to go that route; but if we don't need that much consistency, we can drop that idea.

pjaudiomv:
> There are various style differences with the Users and Login tab tests. How consistent should we be? (Ideally it seems like we would be consistent, but also it isn't essential.)
> 

we should def try and stay somewhat consistent as just easier to read and understand but also dont think we need it to be like a hard rule or whatever

> There is a file `sharedDataAndMocks` for simulated test data like users. I think it would be better to use the shared data file, and add service bodies to that. We'll need the service bodies for some of the users tests, in particular the one that checks for an error message when trying to delete a user that is still associated with a service body.
> 

I updated my tests to use shared data file

> These tests render the ServiceBodyForm directly, whereas the Users tab tests simulate logging in and picking the Users link from the navigation bar. It seems like we should have tests that simulate clicking on the Users and the Service Bodies links, but we don't need all the tests to do that. So rendering the form directly seems better, and I can refactor the Users tests and put the link checking test elsewhere.
> 

I was planning on doing both, these tests mostly just checking form functionality on just the service body form and then adding others to check the service bodies page and click through as well.

> These tests use `fireEvent` to mock clicking and typing. In the other tests I used `@testing-library/user-event` which provides a higher-level API -- I like using `click` and `type` from that package for example.

I was planning on using those for the "click through" tests but can change these to as well, doesnt matter to me.

> The Users and Login tests put all the mocks separate from the tests; and here some of the tests build in the mock (for example `'test Fields are disabled for non-admin users'`). I think the test is cleaner to separate out the mocked method; but I'm sure there is room for debate on this!

I was thinking for a simple mock it wouldnt matter but ya probably best to separate them. I removed that test for now anyway.

> These tests use for example `translations.getString('addServiceBody')` whereas the others just inline the string. I prefer inlining the string, since it tests more of the functionality. That does have the disadvantage that if you change a string used in a test, you have to change it both in localizations and the test; but there aren't that many. Certainly for the language tests, I think we should write
> 
> `await screen.findByText('Ungültiger Benutzername oder Passwort.');`
> 
> rather than `await screen.findByText(translations.getString('invalidUsernameOrPassword');`
> 
> since otherwise we're not checking that we switched to German.

for checking actual translations that totally makes sense. I think I mostly didnt wanna have to go back and change strings but im fine either way.

> So ... if we were trying to make things consistent between the tests, I'd change the tests I wrote to usually render components directly; and I'd change various other things in the ServiceBody test to be like the Users version. I'd be willing to make a consistency pass through all of them if you would like to go that route; but if we don't need that much consistency, we can drop that idea.

if you wanted to make a consistency pass through  im fine with that, im not hard set on doing anything one way. just whatever makes the most sense

alanb2718 (approved):
Looks good now!  We can leave the choice between `fireEvent` and `@testing-library/user-event` be a matter of taste.  Similarly for whether to use `translations.getString` or the actual string, except for the language tests (which do need to use the actual strings).

pjaudiomv:
> Looks good now! We can leave the choice between `fireEvent` and `@testing-library/user-event` be a matter of taste. Similarly for whether to use `translations.getString` or the actual string, except for the language tests (which do need to use the actual strings).

sounds good, thanks Alan

PR #983, opened 2024-07-29
cancel without saving modal

pjaudiomv:
this currently all works except one minor thing.... it resets the form when you hit cancel


alanb2718 (commented):
Nice!!  We do want to not lose the edits if you cancel, but with that change it looks good.

pjaudiomv:
> Nice!! We do want to not lose the edits if you cancel, but with that change it looks good.

I was able to get it to work and added a test for it

alanb2718 (approved):

Issue #1000, opened 2024-08-03
Properly handle miscellaneous errors in the ApiClientWrapper


alanb2718:
The handleErrors method in the ApiClientWrapper has a line at the end of the method for errors that weren't handled earlier:
`return console.log('TODO show error dialog', error.message);`
This is https://github.com/bmlt-enabled/bmlt-root-server/blob/main/src/resources/js/lib/RootServerApi.ts#L291

Change this to handle the error properly and raise an exception.  Right now there are a couple of unit test bugs ([#995](https://github.com/bmlt-enabled/bmlt-root-server/issues/995) and [#999](https://github.com/bmlt-enabled/bmlt-root-server/issues/999)) that were difficult to notice because the test just printed an error message rather than failing.

alanb2718:
I tried replacing this line
`return console.log('TODO show error dialog', error.message);`

with 
`throw new Error('unhandled exception: ' + error.message);`

Then the unit tests fail as expected.  I'm not sure if this is the best way to handle this case, but it does at least catch the error properly.

Also, setting a breakpoint and inspecting the error turns this up:
`cause = TypeError: Invalid URL: //api/v1/auth/token\n`

PR #1002, opened 2024-08-03
fix and move service body modal test

### LINKED ISSUE #995: Fix navbar test for Service Bodies
There seems to be a problem with rendering the Service Bodies tab in the unit tests if you go via the home page and simulate clicking on the link.  See this line, which is currently commented out: https://github.com/bmlt-enabled/bmlt-root-server/blob/main/src/resources/js/tests/Login.spec.ts#L129

If you uncomment it there will be an error message
`TODO show error dialog The request failed and the interceptors did not return an alternative response`
I think it's some interaction between the router, this particular page, and rendering this particular page in unit tests; but I'm not sure.  The existing Service Bodies tests just go straight to the ServiceBodyForm or ServiceBodyModal so they aren't going to encounter this.

This isn't blocking anything, so not urgent to figure it out ... but it's annoying!

### LINKED ISSUE #999: Fix modal dialog test for Service Body Form
There is a bug in this test: https://github.com/bmlt-enabled/bmlt-root-server/blob/main/src/resources/js/tests/ServiceBodyForm.spec.ts#L94

It gives an error message
`TODO show error dialog The request failed and the interceptors did not return an alternative response`
Confusingly, this doesn't actually trigger an error that causes the unit test to fail so it's easy to overlook.

This is similar to this issue: https://github.com/bmlt-enabled/bmlt-root-server/issues/995

pjaudiomv:
this resolves #995 and resolves #999
I think 999 must of been fixed previously. I also thought there could of been some sort of timing issue in login function but couldn't find out. the service bodies page does have to do more things on mount.

### DIFF
--- a/src/resources/js/tests/Login.spec.ts
+++ b/src/resources/js/tests/Login.spec.ts
@@ -123,10 +123,7 @@ describe('navbar tests', () => {
     expect(await screen.findByRole('heading', { level: 2, name: 'Formats' })).toBeInTheDocument();
     replace('/');
     await user.click(await screen.findByRole('link', { name: 'Service Bodies', hidden: true }));
-    // TODO: the following test results in an error message:
-    //     TODO show error dialog The request failed and the interceptors did not return an alternative response
-    // so commented out for now
-    // expect(await screen.findByRole('heading', { level: 2, name: 'Service Bodies',  })).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


alanb2718 (approved):
Looks good!  I'm not seeing the error message with these changes.  We still should fix #1000 -- I have a little more data on that and will add it to the issue.

PR #1013, opened 2024-08-17
Add authenticated endpoint to download laravel log

pjaudiomv:
Should this be restricted to server admin?


alanb2718:
Regarding whether this should be restricted to the server admin: will the log file ever contain information that shouldn't be viewed by a random service body admin, for example information about other service bodies or contacts?  If so, then yes, let's restrict it.  (I'm not sure why anyone other than the server admin would want to download it anyway, but if there isn't confidential information then it doesn't seem like a necessary restriction.)

jbraswell (changes requested):
I think logs/laravel would be better.

Needs swagger annotations so the client can have it.

Needs a failing test with an unauthenticated user and with other user types.

Probably needs policies to enforce permissions?

pjaudiomv:
I changed to logs/laravel , added swagger and restricted to server admin. as well as tests for unauthenticated and service body admin

Screenshot 2024-08-17 at 7 59 54 PM

src/app/Http/Controllers/Admin/LaravelLogController.php
```diff
@@ -0,0 +1,28 @@
...
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Illuminate\Http\Request;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Facades\Gate;
+
+class LaravelLogController extends Controller
+{
+    public function laravelLog(Request $request): JsonResponse|BinaryFileResponse
+    {
+        $user = $request->user();
```
jbraswell, line 15: Permissions should be enforced using a policy. Look at app/Policies.

jbraswell (changes requested):

src/app/Http/Controllers/Admin/LaravelLogController.php
jbraswell: Actually, I think policies only apply if you're dealing with models? It's worth looking into, but your code may be fine as written.

src/app/Http/Controllers/Admin/Swagger/LaravelLogController.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * @OA\Get(path="/api/v1/logs/laravel", summary="Retrieves laravel log", description="Retrieve the laravel log if it exists.", operationId="getLaravelLog", tags={"rootServer"}, security={{"bmltToken":{}}},
+     *     @OA\Response(response=200, description="Returns when user is authenticated.",
+     *         @OA\MediaType(mediaType="text/plain", @OA\Schema(type="string"))
+     *     ),
+     *     @OA\Response(response=401, description="Returns when user is not authenticated.",
+     *         @OA\JsonContent(ref="#/components/schemas/AuthenticationError")
+     *     ),
+     *     @OA\Response(response=403, description="Returns when user is unauthorized to perform action.",
+     *         @OA\JsonContent(ref="#/components/schemas/AuthenticationError")
```
jbraswell, line 19: This should be AuthorizationError.

src/storage/api-docs/api-docs.json
```diff
@@ -486,6 +486,63 @@
...
+                                }
+                            }
+                        }
+                    },
+                    "403": {
+                        "description": "Returns when user is unauthorized to perform action.",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/AuthenticationError"
```
jbraswell, line 523: Should be authorization instead of authentication.

src/app/Http/Controllers/Admin/LaravelLogController.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace App\Http\Controllers\Admin;
+
+use App\Http\Controllers\Controller;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Illuminate\Http\Request;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Facades\Gate;
+
+class LaravelLogController extends Controller
```
jbraswell, line 11: This controller should probably be LogController, to mirror /logs. Then, the method should probably be called laravel, to mirror the "laravel" part of /logs/laravel.

In rest language, logs is the resource type, and laravel is specific instance of that resource.

src/app/Http/Controllers/Admin/LaravelLogController.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $user = $request->user();
+        if (!$user->isAdmin()) {
+            return response()->json(['error' => 'Unauthorized.'], 403);
+        }
+
+        $filePath = storage_path('logs/laravel.log');
+
+        if (file_exists($filePath)) {
+            return response()->download($filePath);
```
jbraswell, line 23: Does laravel have a good way to gzip the outgoing bytes?

jbraswell (changes requested):

src/app/Http/Controllers/Admin/Swagger/LaravelLogController.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @OA\Schema(schema="LaravelLogBase"),
+ */
+class LaravelLogController extends Controller
+{
+    /**
+     * @OA\Get(path="/api/v1/logs/laravel", summary="Retrieves laravel log", description="Retrieve the laravel log if it exists.", operationId="getLaravelLog", tags={"rootServer"}, security={{"bmltToken":{}}},
+     *     @OA\Response(response=200, description="Returns when user is authenticated.",
+     *         @OA\MediaType(mediaType="text/plain", @OA\Schema(type="string"))
```
jbraswell, line 13: Look at the "Response That Returns a File" section of https://swagger.io/docs/specification/describing-responses/.

src/app/Http/Controllers/Admin/LaravelLogController.php
pjaudiomv: Ya I looked into it, it only applies to models

src/app/Http/Controllers/Admin/Swagger/LaravelLogController.php
pjaudiomv: fixed

src/storage/api-docs/api-docs.json
pjaudiomv: fixed

src/app/Http/Controllers/Admin/LaravelLogController.php
pjaudiomv: thanks just altered the things accordingly

src/app/Http/Controllers/Admin/Swagger/LaravelLogController.php
pjaudiomv: fixed, its cool the swagger ui generates a download link for you with the gzip file

src/app/Http/Controllers/Admin/LaravelLogController.php
pjaudiomv: didnt see anything magical with like arrows and shit

src/app/Http/Controllers/Admin/LogController.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $user = $request->user();
+        if (!$user->isAdmin()) {
+            return response()->json(['error' => 'Unauthorized.'], 403);
+        }
+
+        $filePath = storage_path('logs/laravel.log');
+
+        if (file_exists($filePath)) {
+            $fileContent = file_get_contents($filePath);
```
jbraswell, line 22: Rather than reading the whole file into memory, try reading it in chunks. Then you can use streamDownload instead of just download to stream the file back to the client and never have to worry about out of memory errors if the log file is abnormally large. Take a look at the getNawsDump code for an example.

src/app/Http/Controllers/Admin/Swagger/LogController.php
```diff
@@ -0,0 +1,29 @@
...
+     *     ),
+     *     @OA\Response(response=403, description="Returns when user is unauthorized to perform action.",
+     *         @OA\JsonContent(ref="#/components/schemas/AuthorizationError")
+     *     ),
+     *     @OA\Response(response=404, description="Returns when no laravel log file exists.",
+     *         @OA\JsonContent(ref="#/components/schemas/NotFoundError")
+     *     ),
+     * )
+     */
+    public function index()
```
jbraswell, line 26: For consistency this should have the same name as the method in LogController, unless it breaks something.

src/tests/Feature/Admin/LogTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\File;
+
+class LogTest extends TestCase
+{
+    use RefreshDatabase;
+
+    public function testLaravelLogDownloadSuccess()
+    {
+        if (!File::exists(storage_path('logs'))) {
```
jbraswell, line 14: You should avoid mutating anything real in a test. In this case, that's the filesystem. See https://phpunit.de/manual/6.5/en/test-doubles.html#test-doubles.mocking-the-filesystem for information on how to mock the filesystem.

src/tests/Feature/Admin/LogTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $token = $user->createToken('test')->plainTextToken;
+        $this->withHeader('Authorization', "Bearer $token")
+            ->get('/api/v1/logs/laravel')
+            ->assertStatus(200)
+            ->assertHeader('Content-Type', 'text/plain; charset=UTF-8')
+            ->assertHeader('Content-Encoding', 'gzip')
+            ->assertHeader('Content-Disposition', 'attachment; filename=laravel.log.gz');
+    }
+
+    public function testLogDownloadNoLogFailure()
```
jbraswell, line 28: The test name should ideally indicate why it failed. In this case, the name should include NotFound.

jbraswell (changes requested):

src/app/Http/Controllers/Admin/Swagger/LogController.php
pjaudiomv: done

src/tests/Feature/Admin/LogTest.php
pjaudiomv: done

src/app/Http/Controllers/Admin/LogController.php
pjaudiomv: done

src/tests/Feature/Admin/LogTest.php
pjaudiomv: I had tried to use `Storage::fake()` and that didnt work and the other lib they suggested looks like its not maintained anymore so didnt wanna add it as a dep.

jbraswell: I don't feel comfortable approving a test that mutates the filesystem.

pjaudiomv: I commented out the offending two tests for now and added a todo with note

jbraswell (approved):

PR #1015, opened 2024-08-18
add laravel log download link

pjaudiomv:


jbraswell (commented):
I don't think it makes much sense to have this in the account section. There will probably be an admin section, which will also contain a way to upload a NAWS spreadsheet and maybe various other options. It should go there.

Also, I'd use the standard handleErrors stuff that we use everywhere else instead of our try/catch.

pjaudiomv:
not at a place where we need this but this is good example of how to do this, im just gonna close instead of marking wip but keep branch around

PR #1058, opened 2024-09-30
add Formats code

alanb2718:


pjaudiomv (approved):

src/resources/js/components/FormatForm.svelte
```diff
@@ -0,0 +1,257 @@
...
+      <div class="text-gray-700 md:col-span-2 dark:text-gray-300">
+        <strong>{$translations.formatId}:</strong>
+        {selectedFormat?.id}
+      </div>
+    {/if}
+    <div class="md:col-span-2">
+      <Accordion multiple>
+        {#each allLanguages as lang}
+          {#if $translations.getLanguage() === lang}
+            <AccordionItem open>
```
alanb2718, line 174: This code should be simplified to remove the duplication.  Is there a way to make the open argument conditional on `$translations.getLanguage() === lang}`?  If not, the duplication could be removed by factoring out the `AccordionItem` contents into a separate component.

alanb2718: Fixed with #1059

PR #1061, opened 2024-10-02
Auto Geo TZ

pjaudiomv:


alanb2718 (approved):
The Time Zone field should default to something like `Choose option (or leave empty to set the time zone automatically from the location)`

Otherwise looks good!

PR #1063, opened 2024-10-02
Add geocoding class

pjaudiomv:
if no google key set, use Nominatim

https://github.com/bmlt-enabled/bmlt-root-server/blob/pj/geocoding-class/src/resources/js/lib/geocoder.ts#L116-L120

A follow up to this will be selectively loading google maps if key isnt set and including OSM for maps


pjaudiomv:
Do we only want to geocode if one of these fields has changed. Basically if initial values and current values match dont geocode?

```
locationStreet,
locationCitySubsection,
locationMunicipality,
locationProvince,
locationSubProvince,
```

pjaudiomv:
ya I just updated to only geocode if location fields change

pjaudiomv:
wait that wont work for initiall geocode :(

pjaudiomv:
ok fixed that

pjaudiomv:
added the OSM map as well fi googleApiKey unset. So this effectively adds full support for OSM (just dont set google api key)

Issue #1065, opened 2024-10-04
Service Body Full Meeting List editors have "edit" access  to serice bodies


pjaudiomv:
In the new UI, a service body editor (but not owner) can see all the service bodies and it appears they can edit but when saving they get a authorization error (that doesn't bubble up to UI). Also they will never have access to list of service body editors if not user owner so this dropdown may want to be hidden too as it would never be populated.

- Should we just hide the service bodies like we do in old UI
- should we mark all fields read only and hide save button
- other things maybe

pjaudiomv:
resolving closed by https://github.com/bmlt-enabled/bmlt-root-server/pull/1070

PR #1069, opened 2024-10-04
Dont allow saving or editing of servicebodies for non service body ad…

pjaudiomv:
…mins (regular editors)

This attempts to alleviate this issue https://github.com/bmlt-enabled/bmlt-root-server/issues/1065 and removes save button and marks fields readonly if user is only a service body editor and not admin or servicebody admin


alanb2718:
I would just hide the service bodies in this case, as in the old UI.  I've never created users like this so I don't have a use case in mind -- but I'm imagining it would be for situations where you want to give someone the ability to edit meetings in certain service bodies without giving them the credentials for the owner.  In this case, if it's a region, I don't think you would need access to the areas under it.

If others do want to display the service bodies, I'd make the fields read-only so that they appear greyed out.  But the current solution works also.

alanb2718:
ps - problem with this implementation so far: it removes the ability to edit the fields for the service bodies for the owner as well!

pjaudiomv:
Cool I actually think that just hiding the service bodies make sense. Ya we have region volunteers that have access to whole region but none of them are region admins.

pjaudiomv:
closing in favor of https://github.com/bmlt-enabled/bmlt-root-server/pull/1070

PR #1070, opened 2024-10-05
filter service bodies for editors

pjaudiomv:
in support of https://github.com/bmlt-enabled/bmlt-root-server/issues/1065

### DIFF
--- a/src/resources/js/routes/ServiceBodies.svelte
+++ b/src/resources/js/routes/ServiceBodies.svelte
@@ -98,8 +98,15 @@
   $: {
     // prettier-ignore
     filteredServiceBodies = serviceBodies
-      .sort((s1, s2) => s1.name.localeCompare(s2.name))
-      .filter((s) => s.name.toLowerCase().indexOf(searchTerm.toLowerCase()) !== -1);
+          .sort((s1, s2) => s1.name.localeCompare(s2.name))
+          .filter(s =>
+              // Filter based on the search term
+              s.name.toLowerCase().indexOf(searchTerm.toLowerCase()) !== -1 &&
+              // Users that are service body editors but not admins do not have permissions
+              // to perform CUD Actions on Service Bodies, so we hide them.
+              // Ensure that if the user's ID is in assignedUserIds, it's different from adminUserId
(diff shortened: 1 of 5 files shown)


src/resources/js/routes/ServiceBodies.svelte
```diff
@@ -98,8 +98,15 @@
...
-      .sort((s1, s2) => s1.name.localeCompare(s2.name))
-      .filter((s) => s.name.toLowerCase().indexOf(searchTerm.toLowerCase()) !== -1);
+          .sort((s1, s2) => s1.name.localeCompare(s2.name))
+          .filter(s =>
+              // Filter based on the search term
+              s.name.toLowerCase().indexOf(searchTerm.toLowerCase()) !== -1 &&
+              // Users that are service body editors but not admins do not have permissions
+              // to perform CUD Actions on Service Bodies, so we hide them.
+              // Ensure that if the user's ID is in assignedUserIds, it's different from adminUserId
+              (!s.assignedUserIds.includes(<number>$authenticatedUser?.id) || s.adminUserId === $authenticatedUser?.id)
```
alanb2718, line 108: This should be `s.assignedUserIds.includes(<number>$authenticatedUser?.id)`  rather than `!s.assignedUserIds.includes(<number>$authenticatedUser?.id) `

alanb2718 (commented):
There's a bug in the test.  But after I fixed that, it doesn't work ..... the API still gives an error when I try to save the edit.  So either we should change the API to allow this, or just hide the Service Bodies tab entirely as in the old UI.

src/resources/js/routes/ServiceBodies.svelte
pjaudiomv: weird im not seeing same behavior, are you on latest `pj/filter-sbs-editors` branch and running `npm run dev` its working fine for all users for me. The problem with hiding from navbar is we havent loaded service bodies yet when we first need the nav. we previously had decided on lazy loading but im fine with whatever

alanb2718: Hmm .... I just tried pulling from `pj/filter-sbs-editors` branch again, and I am running `npm run dev`. I'm still seeing the same issue.   Not sure what's going on??!!

PR #1071, opened 2024-10-06
add Formats tests

alanb2718:


pjaudiomv (approved):

pjaudiomv:
looks like some sort of error in tests https://github.com/bmlt-enabled/bmlt-root-server/actions/runs/11201858943/job/31137297389

alanb2718:
re the error in the tests: when I run this from the command line, all the tests pass, but there is an error message on the console: 

```
stdout | resources/js/tests/Formats.spec.ts > check content in Formats tab > try to create a format with no translations
Error [ResponseError]: Response returned an error code
    at new ResponseError (/Users/XX/bmlt-root-server/src/node_modules/bmlt-root-server-client/dist/runtime.js:405:28)
    at ApiClient.mockCreateFormat (/Users/borning/bmlt-root-server/src/resources/js/tests/sharedDataAndMocks.ts:778:11)
    at ApiClient.mockCall (file:///Users/XX/bmlt-root-server/src/node_modules/@vitest/spy/dist/index.js:50:17)
    at ApiClient.spy (file:///Users/XX/bmlt-root-server/src/node_modules/tinyspy/dist/index.js:42:80)
    at ApiClientWrapper.createFormat (/Users/XX/bmlt-root-server/src/resources/js/lib/RootServerApi.ts:235:21)
    at onSubmit (/Users/XX/bmlt-root-server/src/resources/js/components/FormatForm.svelte:111:43)
    at HTMLFormElement.handleSubmit (file:///Users/XX/bmlt-root-server/src/node_modules/@felte/core/src/helpers.ts:353:32) {
  response: Response {
    status: 422,
    statusText: 'Unprocessable Content',
    headers: Headers {},
    body: ReadableStream { locked: false, state: 'readable', supportsBYOB: false },
    bodyUsed: false,
    ok: false,
    redirected: false,
    type: 'default',
    url: ''
  }
}
body { message: 'The translations field is required.' }

```
Is this the same thing that is in the log file linked above?  The tests are all passing, but it's weird that this error is showing up on the console.  In production runs, you get that error message for this case (trying to create a format with no translations), and for other errors like trying to log in with an incorrect password.  But on the unit tests, it's only on this unit test that the error is printed to the console.  I was going to put something about this in slack since I don't understand why it's showing up only for this unit test.

PR #1080, opened 2024-10-07
Add support for config array settings using comma seperated strings

pjaudiomv:
allows to set meeting_counties_and_sub_provinces and  meeting_states_and_provinces in docker

IE.  `MEETING_STATES_AND_PROVINCES=MA,NH,VT,ME`

### DIFF
--- a/src/app/LegacyConfig.php
+++ b/src/app/LegacyConfig.php
@@ -103,8 +103,12 @@ class LegacyConfig
         $config['enable_email_contact'] = isset($g_enable_email_contact) && $g_enable_email_contact;
         $config['include_service_body_admin_on_emails'] = isset($include_service_body_admin_on_emails) && $include_service_body_admin_on_emails;
         $config['change_depth_for_meetings'] = $change_depth_for_meetings ?? 0;
-        $config['meeting_states_and_provinces'] = $meeting_states_and_provinces ?? [];
-        $config['meeting_counties_and_sub_provinces'] = $meeting_counties_and_sub_provinces ?? [];
+        $config['meeting_states_and_provinces'] = isset($meeting_states_and_provinces) && is_string($meeting_states_and_provinces)
+            ? collect(explode(',', $meeting_states_and_provinces))->map(fn($id) => trim($id))->toArray()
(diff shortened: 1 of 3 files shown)


src/app/LegacyConfig.php
```diff
@@ -103,8 +103,12 @@ private static function loadConfig()
         $config['enable_email_contact'] = isset($g_enable_email_contact) && $g_enable_email_contact;
         $config['include_service_body_admin_on_emails'] = isset($include_service_body_admin_on_emails) && $include_service_body_admin_on_emails;
         $config['change_depth_for_meetings'] = $change_depth_for_meetings ?? 0;
-        $config['meeting_states_and_provinces'] = $meeting_states_and_provinces ?? [];
-        $config['meeting_counties_and_sub_provinces'] = $meeting_counties_and_sub_provinces ?? [];
+        $config['meeting_states_and_provinces'] = isset($meeting_states_and_provinces) && is_string($meeting_states_and_provinces)
```
pjaudiomv, line 106: We first check if setting is string and then explode on `,` else we default back to setting and either way an empty array if not set or empty sting

src/app/Http/Controllers/Query/SwitcherController.php
```diff
@@ -536,8 +536,12 @@ private function getServerInfo($request): BaseJsonResponse
             'emailEnabled' => legacy_config('enable_email_contact') ? '1' : '0',
             'emailIncludesServiceBodies' => legacy_config('include_service_body_admin_on_emails') ? '1' : '0',
             'changesPerMeeting' => strval(legacy_config('change_depth_for_meetings')),
-            'meeting_states_and_provinces' => implode(',', legacy_config('meeting_states_and_provinces', [])),
-            'meeting_counties_and_sub_provinces' => implode(',', legacy_config('meeting_counties_and_sub_provinces', [])),
+            'meeting_states_and_provinces' => is_array(legacy_config('meeting_states_and_provinces', []))
```
pjaudiomv, line 539: we have to do opposite basically here because ServerInfo outputs as comma separated string

pjaudiomv:
I tested this in the old UI and the new UI as well

Issue #1132, opened 2024-11-10
Consider showing service body selector by default in the Meeting Editor


jbraswell:
Here are some thoughts.

If the user is only assigned one service body, the service body selector shouldn't be shown at all. It may already work like this, I didn't check.

If the user is assigned multiple service bodies, the selector should be shown.

In the case where the selector should be shown, it is likely that users will nearly always want to use it, and having to click the Service Bodies button at the top to show the selector is an extra step and not completely obvious to a first time user. Before the first search runs, there's a lot of unused space where the table goes. It may be good to show the service body selector (and maybe even all the other filters) in the meeting table space until the first search happens. It may also be a good idea to expand the tree structure by default. Clicking through the things is an extra step, and most users won't have access to so many service bodies that it's unruly

PR #1142, opened 2024-11-12
set meeting modal to static height of 95

pjaudiomv:
closes https://github.com/bmlt-enabled/bmlt-root-server/issues/1132


pjaudiomv:
closes https://github.com/bmlt-enabled/bmlt-root-server/issues/1132

PR #1186, opened 2025-04-24
Svelte 5

pjaudiomv:


alanb2718 (approved):
Looks like it's ready!  (How about if we do a major squash and just have a commit message "migrate to svelte 5".)

PR #1198, opened 2025-05-05
tsml format

pjaudiomv:


pjaudiomv:
One potentially awkward part is how I'm mapping formats. Since servers can use arbitrary `key_string` values for formats, I instead map them first to their `worldid_mixed` (which are consistent and unique), and then to TSML type codes. Using the NAWS code (`worldid_mixed`) as an intermediary ensures reliable and consistent mapping to TSML types.

jbraswell (approved):

PR #1205, opened 2025-05-24
WIP: upgrade flowbite to 1.x

pjaudiomv:
Looks like the spinner needs to be fixed and couple other things like the menu, some buttons.  Doesnt look that bad though. This will also simplify some of the configs and upgrade to tailwind 4

```
/workspace/git/bmlt-enabled/bmlt-root-server/src/resources/js/components/SpinnerModal.svelte
  4:7  error  Prefer using writable $derived instead of $state and $effect  svelte/prefer-writable-derived

/workspace/git/bmlt-enabled/bmlt-root-server/src/resources/js/routes/Formats.svelte
  21:7  error  Prefer using writable $derived instead of $state and $effect  svelte/prefer-writable-derived

/workspace/git/bmlt-enabled/bmlt-root-server/src/resources/js/routes/Users.svelte
  19:7  error  Prefer using writable $derived instead of $state and $effect  svelte/prefer-writable-derive
  ```


pjaudiomv:
Ok I totally did have some imports wrong, once i fixed those it fixed a bunch of stuff

alanb2718 (approved):

PR #1206, opened 2025-06-01
rm root

pjaudiomv:
this is the nuclear approach so prob over did it here. the aggregator needs root_server_uri from older servers so prob breaks that but could check for existence of both


jbraswell:
You changed database column names in sql files. Really gotta be careful about a massive search and replace.

jbraswell:
I also wouldn't change class names and variable names at this point. We need a better reason to risk actually breaking something.

jbraswell:
Also please don't change the outputs of our APIs. A rename like this should really only change documentation, repo names, library names, etc.

README.md
```diff
@@ -20,46 +20,46 @@ It is designed to track and locate Narcotics Anonymous meetings, which are regul
...
+For instructions on installing the server, see [installation/README.md](installation/README.md) in the GitHub repository.
 
 OLDER REPOSITORIES
 ------------------
 
 The first BMLT release was in 2009. 
 
 Here are various legacy repositories that are predecessors to the
-[current BMLT Root Server GitHub repository](https://github.com/bmlt-enabled/BMLT-Root-Server).
+[current BMLT Server GitHub repository](https://github.com/bmlt-enabled/BMLT-server).
```
alanb2718, line 54: Probably need to undo the changes to the legacy URLs (or just delete this part -- not sure we still need the links to the really old repos)

alanb2718 (approved):
I noticed a couple of things to fix -- otherwise looks good!

PR #1217, opened 2025-06-12
format things

pjaudiomv:
This would change the Format names on Formats page to include the key. `(BT) Basic Text` however with translations you would end up with `(BT) Basic Text ( translation )` or something like that. I'm not sure we want to do this or may need to alter. Alan I'm guessing you'll have idea or opinion on this.

### DIFF
--- a/src/resources/js/routes/Formats.svelte
+++ b/src/resources/js/routes/Formats.svelte
@@ -101,9 +101,9 @@
     } else {
       const e = format.translations.find((t) => t.language === 'en');
       if (e) {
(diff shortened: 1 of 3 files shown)


pjaudiomv:
Screenshot 2025-06-12 at 10 31 36 AM
Screenshot 2025-06-12 at 10 30 14 AM

maybe its fine actually

alanb2718:
This looks good!  The formats in other languages work just fine -- see screenshot for an example.
german-formats

alanb2718 (approved):

PR #1240, opened 2025-06-23
ensure legacy auth endpoints work with new UI enabled

pjaudiomv:


jbraswell (approved):

jbraswell:
I don't think get_permissions was ever implemented in the rewrite. Do we know what it was supposed to return?

Issue #1250, opened 2025-07-24
Extraneous/Legacy text in the README?


brotskydotcom:
The README in this repo contains these two lines:

```
[Follow this link to access the BMLT Server GitHub repository](https://github.com/bmlt-enabled/bmlt-server).
There are also links to various predecessor legacy repositories [here](#older-repositories) at the end of this README.
```

Their content suggests that this README is also used somewhere else, and that at the end of that usage there is a list of older repositories. But I can't find any other use of this README on the web, so I'm thinking maybe these two lines should just be removed?

I am happy to submit a fix for this should it be deemed a bug and not just a misunderstanding on my part.

pjaudiomv:
I think removing everything below "you can find out more about the bmlt ..." makes sense then after installation add development section 

DEVELOPMENT
------------

For specific information on setting up a development environment for work on the BMLT server, please
see [CONTRIBUTING.md](CONTRIBUTING.md).

PR #1251, opened 2025-07-24
Update README (#1250)

brotskydotcom:
In addition to fixing the issue mentioned in the bug, as per the recommendation there, I made two other changes:

1. I added a LICENSE section after the DEVELOPMENT section.
2. I updated the REQUIREMENTS section to reflect my real-world experience installing from bare metal on a number of different platforms. The changes there were adding info about the required Apache config, database connectivity options, and an additional PHP module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Basic Meeting List Toolbox Server
 
-<h1 align="center">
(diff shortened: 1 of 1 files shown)


jbraswell (approved):
Thank you!

PR #1260, opened 2025-08-06
Correção na rotina de inclusão de formatos

wilsantos:
Correção na tela de inclusão de formatos quando o usuário não informa o campo format_type, que já é definido como default FC1 na linha 469

### DIFF
--- a/src/legacy/local_server/server_admin/c_comdef_admin_ajax_handler.class.php
+++ b/src/legacy/local_server/server_admin/c_comdef_admin_ajax_handler.class.php
@@ -489,7 +489,8 @@ class c_comdef_admin_ajax_handler
 
                     if (!$shared_id) {
                         $shared_id = intval($format_data['shared_id']);
-                        $format_type = $format_data['type'];
+                        if (isset($format_data['type']))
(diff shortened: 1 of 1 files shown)


wilsantos:
> Thank you for finding this bug and this fix! The bug is in the old version of the UI. We have been working on a complete rewrite and plan to release it soon; at that time this code would be removed. Can you articulate reproduction steps for this bug so we can make sure it doesn't exist in the new code? Or if you feel ambitious, you could try it yourself -- here is the current build of the new system: https://s3.amazonaws.com/archives.bmlt.app/bmlt-server/bmlt-server-4.0.0-build993-545ee06bc22d044db1a7a9578f1323aaf7641c6d.zip

Of course, the error occurred when trying to register a new meeting format, if you didn't choose a value for the "NAWS Format" field, it thrown an error and do not save in the database, looking the code, I identified that there was already a treatment to set the default FC1, but it didn't work, i just adjust to it.

alanb2718:
I verified that this bug no longer exists in the new UI code.  Since we will be removing the code for the old UI very soon, I don't think we should make changes to it at this point.  Thanks again for finding this!

PR #1261, opened 2025-08-06
translations download

pjaudiomv:
this downloads translations spreadsheet with english translations, we could update this at somepoint to be able to change the language of translations to download if wanted

Screenshot 2025-08-11 at 2 18 31 PM


alanb2718:
This looks useful!  What do you think about changing it to download the translations file for the currently selected language?  Then also add a help comment saying "Download translations file for the currently selected language" under the button.

The reason for this is that sometimes there is a partially completed translation, which someone might want to revise.  If the language is completely untranslated (i.e., everything is English) then it has the same functionality as at present.

alanb2718:
I did see your original comment about changing the translations language.  That would be another option -- a little more complicated to implement, more flexible although using the login language seems ok.  If you'd like me to change this PR accordingly I could do that.

pjaudiomv:
Ya that makes sense, if you want to change pr go for it!

alanb2718:
ok, will do.

PR #1267, opened 2025-08-16
Handle translations that are nested arrays for sheet export

pjaudiomv:


pjaudiomv:
I think the way it is in PR is easiest for when we go to take a translation and input it but whatever works

PR #1287, opened 2025-09-01
rework default error handlers in ui

jbraswell:


pjaudiomv (approved):
I just took this for a spin, this is some good stuff

PR #1288, opened 2025-09-07
include service body name in meeting response for query api

pjaudiomv:


pjaudiomv:
I understand the concern, but this isn’t something I’m doing lightly. It’s bothered me for eight years, and I’ve raised it a few times. Honestly, it should have been there from the start. Right now, clients often make an extra API call to GetServiceBodies just to get the service body name. Many do it only for that, so this change removes needless round trips. It’s been frustrating to only have the ID without a readable name. Even the Thirsty Fellowships client includes the name, and we should aim higher. The fact we’ve gone a decade without it doesn’t mean we shouldn’t add it now if it clearly adds value. APIs evolve to better serve their consumers. Since this reduces API calls, makes the data model more complete, and fills an obvious gap in design, it seems like a worthwhile improvement

pjaudiomv:
I maintain many different clients myself, so I'm one of the leading consumers of these APIs too. I'm one of the largest customers of the APIs, if you will.

jbraswell:
What features can you build with this feature that couldn't be built without it?

I'm not blocking the change, but I do want to understand if this is just removing a dev annoyance for you or if it's actually enabling new features to end users.

pjaudiomv:
This is primarily removing a dev annoyance, but it translates into real user benefits: faster app loading, reduced data usage, and less server load from fewer API requests. It also eliminates edge cases where the second API call fails while the first succeeds. It’s not about enabling completely new features, it’s about enabling better implementations of existing ones.

jbraswell (approved):

PR #1291, opened 2025-09-13
fallback and show error for google

pjaudiomv:
Currently we don't bubble up geocoding issues, this sends to error modal and then fallback google geocoding to nomintaim


pjaudiomv:
Screenshot 2025-09-13 at 1 49 44 PM

src/resources/js/lib/geocoder.ts
```diff
@@ -1,6 +1,26 @@
 import type { MeetingPartialUpdate } from 'bmlt-server-client';
 import { spinner } from '../stores/spinner';
 import { Loader } from '@googlemaps/js-api-loader';
+import { errorModal } from '../stores/errorModal';
+
+let googleMapsHasErrors = false;
+
+if (typeof window !== 'undefined') {
+  const originalConsoleError = console.error;
+  console.error = function (...args) {
```
jbraswell, line 10: There's got to be a better way to do this than monkey-patching console.error.

pjaudiomv: ya this does smell

PR #1310, opened 2025-10-16
fix passing null value to strlen in GetSearchResults

jbraswell:


jbraswell:
Fixes https://github.com/bmlt-enabled/bmlt-server/issues/1308.

pjaudiomv (approved):

PR #1313, opened 2025-10-19
Allow Meetings to be Efficiently Edited as Groups 

otrok7:
This PR is the purification of an earlier PR, and is intended to allow the ideas here to be considered independently.

The goal of the PR is to allow the maintainer of meeting data for regions where groups have multiple meetings per week to edit the meeting data efficiently, without having to repeat himself.  This is in keeping with the core principle of BMLT, that meeting data should be stored in a single place, the redundancy inevitably leads to inconsistencies.

This PR is NOT about making BMLT group oriented as opposed to meeting oriented.  If crouton or other clients wish to implement a group oriented meeting list, they can do so with or without this PR.

It does not impact regions that do not organize their meetings in this fashion.  It allows both independent meetings and groups to co-exist within the same region.

The basic idea of the implementation is that meetings and groups have very similar data, in particular, both have formats and "meeting_data" (the stuff stored in the _data and _longdata tables).  In this sense, groups are a type of meeting with the following differences:

1.  Groups must not be returned during searches for meetings.
2. Groups do not have a "day" when they meet, and startTime and duration can only be interpreted as default values.
3. Meetings that are part of a group point to the group 

So, the meeting table gets an extra field, allowing participating meetings to point to their group, and another field that marks the entries in the table that represent groups, so that we may easily filter them out from meeting searches.

So, only thing to be done in the backend is to construct the meeting data out of both its own data, and the data associated with its group.  Laravel's ORM makes this relatively easy.

### DIFF
--- a/src/resources/js/components/MeetingEditForm.svelte
+++ b/src/resources/js/components/MeetingEditForm.svelte
@@ -818,9 +818,7 @@
   });
   $effect(() => {
     setData('formatIds', formatIdsSelected);
-    membersOfGroup.forEach((member, i) => {
-      setData('membersOfGroup.' + i + '.formatIds', member.formatIds);
-    });
+    setData('membersOfGroup', membersOfGroup);
   });
   function handleDeleteMember(i: number, setData: (d: any, v: any) => void) {
     if (membersOfGroup.length <= 1) return;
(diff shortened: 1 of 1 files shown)


otrok7:
Changes to UI 1:  Meetings get a button that converts them to groups:
image

Changes to UI 2:  Groups have an additional tab where the member meetings can be edited:
image

PR #1321, opened 2025-10-30
Force PHP to always check for file mods when debugging.

brotskydotcom:
Once PHP loads a file into its opcache, it assumes (by default) that the file won't change for 2 seconds. When I'm running in a docker image, that's too long: I have made a code change and re-request in less than that, and when I do the change doesn't take. Since performance is not paramount when debugging locally, it's safer to force a revalidate (which is just a timestamp check) on every access to a file.


brotskydotcom:
@jbraswell asking you to review as per advice from @pjaudiomv, thanks!

PR #1329, opened 2025-10-31
Correctly map JS 'da' to database 'dk'.

brotskydotcom:
When we look up format names in the database, we have to correct the UI's code for Dansk to match the database's code for Dansk.


brotskydotcom:
Issue description: without this fix, the Formats screen in Danish displays "no translation available" for every format.

jbraswell (approved):

PR #1350, opened 2025-11-06
Fix tests

pjaudiomv:
I think this should work as we set `APP_ENV=testing` in phpunit https://github.com/bmlt-enabled/bmlt-server/blob/main/src/phpunit.xml#L25


pjaudiomv:
ugh it worked locally ha

PR #1364, opened 2025-11-10
Update to newer version of node.

brotskydotcom:
Now that v24 is the latest LTS, we need to be at most one LTS version back, so the latest vite and svelte etc. will work.


pjaudiomv:
> While we're on the topic of build pipelines, I find that I can't do `make test` because my account doesn't have access to the docker repository that has the needed images. Is there some repo I should be requesting access to?

ah the `make test` could prob get a dependency on local docker if you run `make docker` first it should work. all the docker images needed should be public

pjaudiomv:
> No, I just use the latest vite/etc. in other projects and noticed that I couldn't use them here. That forced me to notice that the new node LTS is out so we are no longer on a version that's currently supported by most vendors. So I submitted this just out of general hygiene concerns.
> 
> This PR is a little incomplete, since moving to v22 also changes which packages get installed, and I didn't include the changes to the lock file. I just didn't know what the practice is around these changes in this project, so I went with the minimal version of the change.

weird we are on vite 7 and vitest 4 which are latest

brotskydotcom:
> ah the `make test` could prob get a dependency on local docker if you run `make docker` first it should work. all the docker images needed should be public

That did the trick, thanks!! OK for me to have`test` rely on `docker`?

PR #1375, opened 2025-11-15
Add state persistence

pjaudiomv:
Adds Svelte stores to persist state when switching between editors (formats, meetings, service bodies, users). Users now retain their search filters, scroll position, and last edited item when navigating between tabs.  This has added benefit of making UI seem snappier not having to re-fetch for every page load.

resolves https://github.com/bmlt-enabled/bmlt-server/issues/1134


pjaudiomv:
closes https://github.com/bmlt-enabled/bmlt-server/issues/1134

PR #1389, opened 2025-11-20
migrate server settings to db

pjaudiomv:
Currently deployed to unstable here https://unstable.aws.bmlt.app/main_server/#/administration

Screenshot 2025-11-20 at 3 42 11 PM

### DIFF
--- a/src/resources/js/components/TagInput.svelte
+++ b/src/resources/js/components/TagInput.svelte
@@ -22,13 +22,13 @@
       .map((item) => item.trim())
       .filter((item) => item && !value.includes(item));
     if (items.length > 0) {
-      value = [...value, ...items];
+      value = [...value, ...items].sort();
       inputValue = '';
     }
   }
 
   function removeItem(index: number) {
-    value = value.filter((_, i) => i !== index);
+    value = value.filter((_, i) => i !== index).sort();
   }
 
   function handleKeydown(event: KeyboardEvent) {


src/config/app.php
```diff
@@ -106,7 +106,8 @@
     |
     */
 
-    'locale' => legacy_config('language', 'en'),
+    // TODO: Load from settings table once we resolve bootstrap ordering
+    'locale' => 'en', // legacy_config('language', 'en'),
```
pjaudiomv: app locale is used here

```
14:21:39 › grep -lr "locale" ./
./Repositories/MeetingRepository.php
./Http/Controllers/Query/SwitcherController.php
```

docker/start-bmlt.sh
```diff
@@ -1,6 +1,16 @@
...
 
-/bin/bash /tmp/write-config.sh
+echo "Waiting for database..."
+until /usr/bin/php /var/www/html/main_server/artisan db:show >/dev/null 2>&1; do
+  sleep .2
+done
+echo "connected"
+
+/usr/bin/php /var/www/html/main_server/artisan migrate --force
+/usr/bin/php /var/www/html/main_server/artisan settings:sync --force
```
jbraswell: If the migrations are already syncing the settings, can this line be removed?

src/app/Http/Controllers/Admin/SettingController.php
```diff
@@ -0,0 +1,66 @@
...
+    private SettingRepositoryInterface $settingRepository;
+
+    public function __construct(SettingRepositoryInterface $settingRepository)
+    {
+        $this->settingRepository = $settingRepository;
+    }
+
+    public function index(Request $request)
+    {
+        $this->authorize('viewAny', Setting::class);
```
jbraswell: You should be able to call `$this->authorizeResource` in the controller rather than having to call `$this->authorize` in each endpoint. See the other admin api controllers.

src/app/Http/Controllers/Admin/SettingController.php
```diff
@@ -0,0 +1,66 @@
...
+        $validated = validator($request->all(), [
+            '*' => function ($attribute, $value, $fail) {
+                if (!array_key_exists($attribute, Setting::SETTING_TYPES)) {
+                    $fail("The setting key '{$attribute}' is not valid.");
+                }
+            },
+        ])->validate();
+
+        $rules = [];
+        foreach ($validated as $key => $value) {
```
jbraswell: You're running a validator against the payload twice. Once to validate the keys, and again to validate the types. You should be able to just do it all in one pass.

src/app/Http/Controllers/Query/SwitcherController.php
```diff
@@ -357,7 +357,7 @@ private function getSearchResults(Request $request, ?string $dataFormat = null):
 
         // This code to calculate the formats fields is really inefficient, but necessary because
         // we don't have foreign keys between the meetings and formats tables.
-        $langEnum = $request->input('lang_enum', config('app.locale'));
+        $langEnum = $request->input('lang_enum', legacy_config('language') ?: config('app.locale'));
```
jbraswell: Why is this change needed?

src/app/Http/Controllers/UserInterfaceController.php
```diff
@@ -42,8 +38,8 @@ public static function handle(Request $request): Response
             'isLanguageSelectorEnabled' => legacy_config('enable_language_selector'),
             'languageMapping' => self::getLanguageMapping(),
             'formatLangNames' => legacy_config('format_lang_names', []),
-            'meetingStatesAndProvinces' => implode(',', legacy_config('meeting_states_and_provinces', [])),
-            'meetingCountiesAndSubProvinces' => implode(',', legacy_config('meeting_counties_and_sub_provinces', [])),
+            'meetingStatesAndProvinces' => legacy_config('meeting_states_and_provinces', []),
+            'meetingCountiesAndSubProvinces' => legacy_config('meeting_counties_and_sub_provinces', []),
```
jbraswell: Do any clients depend on this being a comma delimited list? Does BMLT Workflow use this?

src/app/Models/Setting.php
```diff
@@ -0,0 +1,405 @@
...
+        'includeServiceBodyEmailInSemantic' => false,
+        'bmltTitle' => 'BMLT Administration',
+        'bmltNotice' => '',
+        'formatLangNames' => [],
+    ];
+
+    /**
+     * Get the value attribute with JSON decoding.
+     */
+    public function getValueAttribute($value)
```
jbraswell: Where is this used?

src/app/Models/Setting.php
```diff
@@ -0,0 +1,405 @@
...
+                return (string) $value;
+        }
+    }
+
+    /**
+     * Get typed value for a setting key.
+     */
+    public function getTypedValue()
+    {
+        return $this->value;
```
jbraswell: Is this method really necessary if all it does is return `$this->value`?

jbraswell (changes requested):

docker/start-bmlt.sh
pjaudiomv: no because the `settings:sync` command will read from environment variables, as we are removing the `[write-config.sh](https://github.com/bmlt-enabled/bmlt-server/blob/main/docker/write-config.sh)` we need this because of that too

src/app/Http/Controllers/Admin/SettingController.php
pjaudiomv: ok cool ill check that out

src/app/Models/Setting.php
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    /**
+     * Sync settings from environment variables to database.
+     * Reads from: 1) environment variables, 2) defaults.
+     *
+     * @param bool $force If false, skip settings that already exist in database
+     * @return array ['synced' => [...], 'skipped' => [...], 'actions' => [...]]
+     */
+    public static function syncFromEnvironment(bool $force = false): array
```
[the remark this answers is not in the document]
pjaudiomv: we need a way to alter server behavior from environment variables, i was unsure how else to do this and have frontend api be happy as well

src/app/Models/Setting.php
pjaudiomv: no this can be removed and is left over from before i moved the migration logic into migration

docker/auto-config.inc.php
```diff
@@ -17,23 +15,9 @@
...
-
-$include_every_admin_on_emails = FALSE; // If this is TRUE (or 1), then any emails sent using the meeting contact will include all Service Body Admin contacts (including the Server Administrator) for the meeting (ignored, if $g_enable_email_contact or $include_service_body_admin_on_emails is FALSE).
-
-
-$time_format = 'g:i A'; // The PHP date() format for the times displayed.
-$change_date_format = 'g:i A, n/j/Y'; // The PHP date() format for times/dates displayed in the change records.
-$admin_session_name = 'BMLT_Admin'; // This is merely the 'tag' used to identify the BMLT admin session.
-$default_minute_interval = 5; // This sets the minutes interval for Start Time and Duration.
+$g_include_service_body_email_in_semantic = FALSE; //Set this to TRUE (or 1) to include including Service body contact emails in the semantic response
+$g_defaultClosedStatus = TRUE;   // If this is FALSE (or 0), then the default (unspecified) Open/Closed format for meetings reported to NAWS is OPEN. Otherwise, it is CLOSED.
```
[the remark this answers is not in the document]
pjaudiomv: I had it left in for testing

PR #1394, opened 2025-11-24
Add download spreadsheet button to users and servicebodies for server admin

pjaudiomv:
Screenshot 2025-11-24 at 1 17 12 PM

closes https://github.com/bmlt-enabled/bmlt-server/issues/391


pjaudiomv:
> This looks good! The only change that might be useful would be to also test the contents of the downloaded spreadsheets, as was done with the Administration tests. I'd be willing to add those if you would like.

Ya sure go for it, thanks 🙏

PR #1405, opened 2025-11-29
New format schema

otrok7:
Still some work to do on tests: e.g. the ones generate formats directly in the DB need to be adjusted.  But I thought you should maybe review in case something needs to be adjusted for aggregator.  For instance, the pivot table doesn't have an ID column, it uses the meeting and format ID combined as the PK...I hope aggregator is OK with this...

The code to save changes doesn't really work yet:  these are never shown in the UI, maybe it's better to remove the feature?  Up to you...

This doesn't clean up the code as much as I hoped, because we never actually hydrate the relationship, we always return two tables, connected by PKs.  And even if we didn't have the old API to force this, JSON probably forces it.

### DIFF
--- a/src/app/Models/FormatTranslation.php
+++ b/src/app/Models/FormatTranslation.php
@@ -31,15 +31,4 @@ class FormatTranslation extends Model
     {
         return $this->belongsTo(FormatMain::class, 'shared_id_bigint', 'shared_id_bigint');
     }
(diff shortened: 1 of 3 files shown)


otrok7:
The remaining tests seem to be connected with aggregator, which I think you guys wanted to do.

PR #1407, opened 2025-12-02
Add frontend translation helper cmds

pjaudiomv:
```
> GITHUB_ACTIONS=true php artisan translation

Available commands for the "translation" namespace:
  translation:add                      Add a translation key/value to all language files in alphabetical order
  translation:delete                   Delete a translation key from all language files
  translation:update-from-spreadsheet  Update translation file from XLSX spreadsheet
```

### artisan translation:add
```php
# Add a new key
php artisan translation:add meetingName "Meeting Name"

# Update an existing key (with --force)
php artisan translation:add cancel "Cancelled" --force
```

### artisan translation:delete
```php
php artisan translation:delete oldKey
```

### artisan translation:update-from-spreadsheet
```php
php artisan translation:update-from-spreadsheet /path/to/italian-translations.xlsx it
```

**Note:** if you try to test this with main, currently you'll need to either have env var set or prefix with `GITHUB_ACTIONS=true`

### DIFF
--- a/src/app/Console/Commands/AddTranslationKey.php
+++ b/src/app/Console/Commands/AddTranslationKey.php
@@ -1,0 +1,233 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\File;
+
+/**
+* # Add a new key
+* php artisan translation:add meetingName "Meeting Name"
+*
+* # Update an existing key (with --force)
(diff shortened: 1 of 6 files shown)


alanb2718 (approved):
Looks good!  One minor issue: if you add a key say `aaaaTest` that would be added at the beginning of the list, it puts it on the same line as 
`export const enTranslations = {`

Then if you run the prettier, it will put it on its own line but with a blank line afterwards.

PR #1413, opened 2025-12-14
Ab/fix format migrations

alanb2718:


jbraswell (approved):
Needs to be rebased before merging to ensure the latest migration is still referenced in the middleware.

PR #1414, opened 2025-12-14
changelog fix

pjaudiomv:


alanb2718 (commented):
Yes, we should change this .... but might as well wait and also add a release date and tag to the changelog??

PR #1417, opened 2025-12-19
move USER_AGENT into a constant at top of file for better visibility

alanb2718:


jbraswell (approved):

src/app/Console/Commands/ImportRootServers.php
```diff
@@ -28,6 +28,8 @@
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Http;
 
+const USER_AGENT = 'Mozilla/5.0 (X11; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0 +aggregator';
```
pjaudiomv: should we update this to lil more modern while we are here

`'Mozilla/5.0 (X11; Linux x86_64; rv:146.0) Gecko/20100101 Firefox/146.0 +aggregator';`

PR #1428, opened 2025-12-27
fix danish lang code

pjaudiomv:
This does two things

1. Fixes the Danish language code which should be `da` (`ISO 639-1`) not `dk` (country code, `ISO 3166-1 alpha-2`).

2. Replaces `Locale::getDisplayLanguage()` with translation files for language names, eliminating the need for PHP intl extension which has been pain point with some servers.


src/tests/Feature/GetFieldKeysTest.php
```diff
@@ -102,7 +102,7 @@ public function testGetFieldKeysItalian()
     public function testGetFieldKeysAllLocales200()
     {
         try {
-            $locales = ['de', 'dk', 'en', 'es', 'fa', 'fr', 'it', 'pl', 'pt', 'ru', 'sv'];
+            $locales = ['de', 'da', 'en', 'es', 'fa', 'fr', 'it', 'pl', 'pt', 'ru', 'sv'];
```
alanb2718: fix alphabetization ....

alanb2718 (approved):
Should add migrations for `comdef_service_bodies` and `comdef_users`.  With those changes looks good to go!

PR #1436, opened 2026-01-04
Add admin notes field

pjaudiomv:
This would close https://github.com/bmlt-enabled/bmlt-server/issues/450 and I think we could also close https://github.com/bmlt-enabled/bmlt-server/issues/82 as it creates a way to accomplish it abeit requires user training 

Screenshot 2026-01-04 at 6 26 49 PM


alanb2718 (approved):
Looks good!

PR #1440, opened 2026-01-06
Add resource IDs

pjaudiomv:
Screenshot 2026-01-06 at 5 19 12 PM
Screenshot 2026-01-06 at 5 16 51 PM
Screenshot 2026-01-06 at 5 16 41 PM
Screenshot 2026-01-06 at 9 49 27 AM


pjaudiomv:
> This looks great! It would be good to have some unit tests for this though. I'd be willing to write some if you would like.

Just added tests!

PR #1486, opened 2026-04-09
fix test for reserved formats to work for translations with diff keys

alanb2718:


pjaudiomv (approved):
nice this wont catch THY or VMA etc anymore either

PR #1494, opened 2026-04-19
add Laravel Boost MCP dev tooling

pjaudiomv:
Adds [Laravel Boost](https://laravel.com/ai/boos) [Docs](https://laravel.com/docs/13.x/boost) as a dev dependency, which provides an MCP server with tools for AI assist dev work. 

- `boost.json` - configures which Boost features are enabled for this project
- `.mcp.json` - wires the Boost MCP server to run via docker exec into our existing docker-bmlt-1 container

AI Flavors will change im sure so didnt want to add anything specific to any one, thats why just added boost which works with all of them.

I stopped short of adding a `.claude/settings.json` but if your using claude for frontend should def add svelte plugin or this in your `.claude/settings.json` 

https://svelte.dev/docs/ai/overview

```json
22:45:10 › cat .claude/settings.json
{
  "enabledPlugins": {
    "svelte@svelte": true
  }
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,9 @@ $(ZIP_FILE): $(VENDOR_AUTOLOAD) $(FRONTEND) $(SEMANTIC_HTML) $(TIMEZONE_ASSETS)
 		-x main_server/.nvmrc \
 		-x main_server/.phpcs.xml \
 		-x main_server/.phpstan.neon \
+		-x main_server/.mcp.json \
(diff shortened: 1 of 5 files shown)


pjaudiomv:
Alternatively we can just add `boost.json` and `.mcp.json` to gitignore and let devs manage their own settings and just add the one dev dependency

PR #1495, opened 2026-04-23
check for valid timezone data response

pjaudiomv:


alanb2718 (approved):
Please fix the issue regarding the error message -- otherwise looks great!

Regarding the error message: when I artificially cause an error, the UI puts an appropriate message under "Time Zone" (good), but then at the bottom of the screen it says "Tabs with one or more errors: Basic".  It should say "Location" instead.

And a totally picky point (fee free to ignore this): for the English version of `timeZoneGeocodeError`: add a period at the end so that it is `'Unable to determine timezone from coordinates. Please select a timezone manually.'`

Otherwise the first part is a sentence and the second isn't.  (Do this also for the other untranslated errors; most of the translated errors are OK.)

PR #1499, opened 2026-04-29
add /servicebodies/{id}/editors endpoint and surface hidden editors in form

### LINKED ISSUE #1488: Preserve hidden service body editors when non-admin updates a service body
TL;DR Service body update removes editors not visible to service body admin

When a service body admin edits their service body, the `GET /api/v1/users` endpoint only returns users they own (i.e., users with `owner_id_bigint` matching their own ID). If the service body had editors assigned that weren't owned by that admin, those editors wouldn't appear in their users list. The frontend ([ServiceBodyForm.svelte](https://github.com/bmlt-enabled/bmlt-server/blob/main/src/resources/js/components/ServiceBodyForm.svelte#L63)) correctly filters the selection to only visible users, but shows save button as active and form `isDirty`,  when saved, the invisible editors were simply absent from the submission, causing them to be wiped.

pjaudiomv:
endpoint returns `{userId, displayName, readOnly}` tuples. `readOnly` is true when the caller can't manage that user via the users API. Only display names are exposed, no emails or usernames. `ServiceBodyForm` fetches this on edit and renders an "Other Meeting Editors" read-only badge list under the multi-select. The `hiddenUserIds` fallback from #1497 is preserved. Closes #1488

Screenshot 2026-04-29 at 4 43 24 PM


alanb2718 (approved):
Looks great!

PR #1517, opened 2026-07-26
Fix blur focused field on form submit to prevent reactivity lockup

pjaudiomv:


alanb2718 (approved):
Thanks for fixing this!

PR #1520, opened 2026-08-03
Pj/fix mysql fulltext search

### LINKED ISSUE #1519: Meeting that meets on two days only appears once in the BMLT Admin & Semantic Server. But does appear correctly with Crouton
We discovered that after the update to BMLT 4.2.5, searching on this meeting in the BMLT Admin only brings up the Monday meeting. They also meet on Friday, Recovery Has No Limits

Crouton on naworks.org:
Image

BMLT ADMIN panel:
Image

Searching on the group name in the Semantic Wizard only bring up the Monday meeting:
https://meetings.naworks.org/main_server/client_interface/json/?switcher=GetSearchResults&venue_types=1&SearchString=%22Recovery%20Has%20No%20Limits%22

Image

We tried rolling back to 4.2.3, but it still only shows the Monday listing. We thought maybe the names had a sight difference or hidden character, so copied the Monday name into Friday. They must have been identical because the "Apply Changes" remained greyed out.

If you search on their meeting ID, placing: 1204,1205 into the search field does bring up both meetings.

Image

If you need more info, let us know.

pjaudiomv:
## Summary

Text searches silently dropped meetings on MySQL. Searching "Recovery Has No Limits" returned only the Monday meeting (1204) and not the Friday one (1205), even though both exist and are individually retrievable by ID.

This was not a 4.2.5 regression — the rollback to 4.2.3 in the report didn't help because the bug predates both.

Fixes #1519

## Root cause

`MeetingRepository::getSearchResults` built the text search as two correlated `EXISTS` subqueries (`whereHas('data')->orWhereHas('longdata')`), each containing `MATCH() AGAINST()`.

MySQL plans that as a `DEPENDENT SUBQUERY` with `type: fulltext` / `Ft_hints: sorted`, and returns wrong results — the fulltext scan is re-evaluated per outer row and rows go missing. The tell, against the reporting server's data:

- query meeting 1205 alone → found
- query meeting 1204 alone → found
- query both together → **only 1204**

MariaDB executes the same plan correctly, so this was invisible to dev and CI, which run MariaDB only.

## Impact

Not one meeting — every text search on MySQL lost results. Against the reporting server's database on MySQL 8.4:

| search term | correct | returned | lost |
|---|---|---|---|
| Recovery Has No Limits | 81 | 69 | 15% |
| Church | 270 | 202 | 25% |
| Philadelphia | 216 | 159 | 26% |
| Street | 171 | 135 | 21% |
| New Beginnings | 80 | 61 | 24% |
| How It Works | 19 | 14 | 26% |

Every dropped meeting shared a field value with another meeting. Verified as pure false negatives — no wrong rows returned, just missing ones.

## Fix

Rewrite the two clauses as uncorrelated `whereIn` subqueries, so the fulltext scan runs once instead of once per candidate meeting.

Set-level comparison against ground truth on MySQL 8.4 ("Church"): truth 270, before 202, after 270 — exact match, no false positives, no false negatives. Holds under added filters (published, weekday, service body, `ORDER BY`/`LIMIT`).

Also ~20x faster on MySQL, since the correlated plan was the source of both problems:

| term | before | after |
|---|---|---|
| Church | 68.3 ms | 3.5 ms |
| Philadelphia | 121.1 ms | 4.0 ms |
| Street | 58.7 ms | 3.6 ms |

## MariaDB is unaffected

Identical results before and after on 10.2, 10.11, 11.4, 11.8 and 12.3, and identical timing (12.3: 80.4 ms → 80.1 ms). MariaDB's numbers already matched what MySQL now returns — the fix makes MySQL agree with MariaDB, it doesn't change MariaDB.

Also checked at 20x scale (10,140 meetings / 138,760 data rows): counts identical, timings within noise on both engines.

## Tests

- `testSearchStringMatchesDuplicateValues` — behavioural, meetings sharing a name all return.
- `testSearchStringFullTextSubqueriesAreNotCorrelated` — the real guard.

The shape assertion is deliberate. Whether MySQL actually drops rows depends on the plan it picks, which varies with table statistics — at test-suite scale it picks a semi-join and returns correct results, so a behavioural test can't reproduce this reliably, and wouldn't fail on MariaDB regardless. The shape test asserts the `MATCH()` subqueries don't reference the outer table; it fails without the fix on MariaDB, so it genuinely guards the regression in CI.

## CI

Added `mysql:8.4` to the test matrix alongside `mariadb:10.2` and `mariadb:latest` (6 → 9 jobs). The service block now takes a full image ref so one definition covers both engines; `MYSQL_*` env vars are accepted by both.

closes https://github.com/bmlt-enabled/bmlt-server/issues/1519

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,19 +18,20 @@ jobs:
           - "8.4"
           - "8.5"
         db:
-          - "10.2"
-          - "latest"
+          - "mariadb:10.2"
+          - "mariadb:latest"
+          - "mysql:8.4"
     services:
-      mariadb:
-        image: mariadb:${{ matrix.db }}
+      db:
+        image: ${{ matrix.db }}
         ports:
           - 3306
         env:
-          MARIADB_USER: rootserver
-          MARIADB_PASSWORD: rootserver
-          MARIADB_DATABASE: rootserver
-          MARIADB_ROOT_PASSWORD: rootserver
-        options: --health-cmd="${{ (matrix.db == 'latest') && 'mariadb-admin' || 'mysqladmin' }} ping" --health-interval=5s --health-timeout=2s --health-retries=3
+          MYSQL_USER: rootserver
+          MYSQL_PASSWORD: rootserver
+          MYSQL_DATABASE: rootserver
+          MYSQL_ROOT_PASSWORD: rootserver
+        options: --health-cmd="${{ (matrix.db == 'mariadb:latest') && 'mariadb-admin' || 'mysqladmin' }} ping" --health-interval=5s --health-timeout=2s --health-retries=3
     steps:
       - name: checkout 🛒
         uses: actions/checkout@v6
@@ -52,12 +53,12 @@ jobs:
 
       - name: make test 🧪
         env:
-          DB_PORT: ${{ job.services.mariadb.ports[3306] }}
+          DB_PORT: ${{ job.services.db.ports[3306] }}
         run: |
           make test
 
       - name: Send coverage data to codecov.io 📀
-        if: matrix.db == 'latest' && matrix.php == '8.4'
+        if: matrix.db == 'mariadb:latest' && matrix.php == '8.4'
         uses: codecov/codecov-action@v5
         with:
           files: src/coverage.xml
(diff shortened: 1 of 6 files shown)


pjaudiomv:
Not that this issue has been reported multiple times in the passed and we never could reproduce because it was mysql specific issue not mariadb

jbraswell (approved):

